PR #6, opened 2021-10-14
Add plotting tools

elynnwu:
- Add model output in the example run file
- Add plotting script for visualizing Python output vs. Fortran output


FlorianDeconinck (approved):
LGTM

PR #9, opened 2021-10-26
Add script to check if package or its dependencies have changed for tests

mcgibbon:
Changes:
- Add changed_from_main.py which prints "true" or "false" based on whether a project or its dependencies have changed compared to the `main` branch
- Update `make test_util` target to only run if it has changed compared to the main branch
- Update dotfile to use labels equivalent to subdirectory names, as this file is now used to as the reference for dependencies in changed_from_main.py


changed_from_main.py
```diff
@@ -0,0 +1,89 @@
+"""
+This script determines whether one of the projects in the repo or any of its
```
twicki, line 2: should this really be in the top level directory? 

In fv3core we did use a `.jenkins` folder for these type of things as it is pretty CI related. Or we could put this into a util folder?

changed_from_main.py
```diff
@@ -0,0 +1,89 @@
...
+This is useful for running tests only on projects that have changed.
+
+This script should depend only on Python 3.6+ standard libraries.
+"""
+import argparse
+import re
+import os
+import subprocess
+from typing import Dict, Sequence
+import sys
```
twicki, line 14: sys seems to be unused

changed_from_main.py
```diff
@@ -0,0 +1,89 @@
...
+            subdir = name_to_subdir[name]
+            dependency_subdir = name_to_subdir[dependency_name]
+            subdir_dependencies[subdir].append(dependency_subdir)
+    return subdir_dependencies
+
+SUBDIR_DEPENDENCIES = get_dependencies()
+
+
+def parse_args():
+    # we use 1 for changed so that if something goes wrong and there is an
```
twicki, line 45: why is this comment under the parsing?

changed_from_main.py
```diff
@@ -0,0 +1,89 @@
...
+    for groups in DEPENDENCY_PATTERN.findall(dotfile_text):
+        name = groups[0]
+        dependency_name = groups[1]
+        if name in name_to_subdir:
+            subdir = name_to_subdir[name]
+            dependency_subdir = name_to_subdir[dependency_name]
+            subdir_dependencies[subdir].append(dependency_subdir)
+    return subdir_dependencies
+
+SUBDIR_DEPENDENCIES = get_dependencies()
```
twicki, line 41: the reason this is globally defined is because it is used in the argparse, right? 

would it nevertheless be cleaner if we call `get_dependencies` first and pass this into parse args? It feels slightly unnatural to have this be defined globally.

dependencies.dot
```diff
@@ -1,14 +1,16 @@
...
+# each folder entry must have a "label" equal to its directory name
 digraph {
-  Pace [shape=box]
-  fv3core [shape=oval]
-  driver [shape=oval]
-  physics [shape=oval]
-  util [shape=oval]
+  pace [shape=box]
+  fv3core [shape=oval, label="fv3core"]
+  driver [shape=oval, label="driver"]
```
twicki, line 6: driver does not exist yet. should it already be here?

changed_from_main.py
```diff
@@ -0,0 +1,89 @@
+"""
```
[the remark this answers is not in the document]
mcgibbon: Updated.

changed_from_main.py
mcgibbon: If our top level directory gets cluttered I could see moving this eventually, but this is an integral part of running the top-level Makefile targets as we're currently planning to use it. If we call it directly in Jenkins instead of calling it inside the Makefile (as I did for test_util), we can move it then.

Makefile
```diff
@@ -6,6 +6,7 @@ CWD=$(shell pwd)
 PULL ?=True
 CONTAINER_ENGINE ?=docker
 RUN_FLAGS ?=--rm
+CHECK_CHANGED_SCRIPT=python changed_from_main.py
```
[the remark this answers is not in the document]
mcgibbon: I've updated it to use a shebang, and committed the chmod +x change to the file.

Makefile
```diff
@@ -18,16 +19,18 @@ dev:
 		$(FV3GFS_IMAGE) bash
 
 test_util:
-	$(MAKE) -C fv3gfs-util test
+	if [ $(shell $(CHECK_CHANGED_SCRIPT) fv3gfs-util) = true ]; then \
```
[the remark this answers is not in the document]
mcgibbon: Updated.

Makefile
```diff
@@ -18,16 +19,18 @@ dev:
...
+		$(MAKE) -C fv3gfs-util test; \
+	fi
 	
 savepoint_tests:
 	$(MAKE) -C fv3core savepoint_tests
 
 savepoint_tests_mpi:
 	$(MAKE) -C fv3core savepoint_tests_mpi
```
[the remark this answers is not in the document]
mcgibbon: Updated.

changed_from_main.py
mcgibbon: Fixed. I've gotten to a point of assuming all my imports are used because pre-commit will catch it, and forgot pre-commit isn't set up yet.

changed_from_main.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+DIRNAME = os.path.dirname(os.path.abspath(__file__))
+DEPENDENCIES_DOTFILE = os.path.join(DIRNAME, "dependencies.dot")
+
+DEFINITION_PATTERN = re.compile(r"\s*([a-zA-Z0-9]+) \[.*label=\"(.*)\".*\]")
+DEPENDENCY_PATTERN = re.compile(r"\s*([a-zA-Z0-9]+) -> ([a-zA-Z0-9]+)")
+
+
+def get_dependencies() -> Dict[str, Sequence[str]]:
```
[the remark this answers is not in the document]
mcgibbon: Fixed.

changed_from_main.py
mcgibbon: It would have made more sense why if it were actually accurate, but this is a left-over comment. Deleted.

changed_from_main.py
mcgibbon: Good idea. Updated.

dependencies.dot
mcgibbon: No harm in including it because it's not depended on by any other components, it can happen now or when the folder gets added. I believe it will be added as part of Rhea's current PR(s).

dependencies.dot
```diff
@@ -1,14 +1,16 @@
...
+  physics [shape=oval, label="fv3gfs-physics"]
+  util [shape=oval, label="fv3gfs-util"]
   
-  Pace -> fv3core
-  Pace -> physics
-  Pace -> util
-  Pace -> driver
+  pace -> fv3core
+  pace -> physics
+  pace -> util
```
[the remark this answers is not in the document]
mcgibbon: All tests which need to run are run simultaneously. These dependencies are not used for testing order and do not imply anything w.r.t. testing order.

dependencies.svg
```diff
@@ -4,15 +4,15 @@
 <!-- Generated by graphviz version 2.49.2 (20211016.1639)
  -->
 <!-- Pages: 1 -->
-<svg width="230pt" height="260pt"
- viewBox="0.00 0.00 230.24 260.00" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg width="287pt" height="260pt"
```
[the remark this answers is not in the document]
mcgibbon: Added a note along these lines. I'd usually prefer a pre-commit hook, but I would like to avoid breaking pre-commit on systems without `dot` installed.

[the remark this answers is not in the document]
twicki: I think that makes a lot of sense

changed_from_main.py
twicki: sounds good to me

dependencies.dot
twicki: it's fine if it is here. Right now it exits not very gracefully if you call `changed_from_main.py driver` but this is very temporary. If it is not blocking any CI related work this is totally fine!

twicki (commented):
lgtm

twicki (approved):
lgtm

PR #15, opened 2021-11-12
Merge submodules as subtrees

mcgibbon:
This PR merges files from fv3gfs-util, fv3gfs-physics, and fv3core as subtrees instead of submodules. Nested submodules are moved to the top level `external` directory to ensure consistency, and because submodules can't exist inside subtrees. These and other top-level cross-dependencies are accessed using symbolic links.


elynnwu (approved):
Thanks so much for getting the tests to pass!

PR #19, opened 2021-11-16
Allow objects with storage attribute as call arguments to FrozenStencil

mcgibbon:
This PR modifies FrozenStencil so that it checks for a `.storage` attribute and uses that instead when called.

### DIFF
--- a/fv3core/fv3core/stencils/dyn_core.py
+++ b/fv3core/fv3core/stencils/dyn_core.py
@@ -594,12 +594,12 @@ class AcousticDynamics:
                 )
                 self.riem_solver_c(
                     dt2,
-                    state.cappa,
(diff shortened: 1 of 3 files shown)


fv3core/fv3core/stencils/dyn_core.py
```diff
@@ -594,12 +594,12 @@ def __call__(self, state):
                 )
                 self.riem_solver_c(
                     dt2,
-                    state.cappa,
+                    state.cappa_quantity,
```
mcgibbon, line 597: This change is done mainly to show that the code works, it will be swiftly edited anyways when all of these are passing Quantity objects.

eddie-c-davis (approved):
LGTM!

PR #24, opened 2021-11-19
Add jenkins at top level

elynnwu:
- Adding `.jenkins` at the top level for setting up pace level environment

### DIFF
--- a/.jenkins/actions/lint.sh
+++ b/.jenkins/actions/lint.sh
@@ -3,7 +3,7 @@ set -e -x
 env_name=venv-${BUILD_NUMBER:-0}
 python3 -m venv ${env_name}
 . ${env_name}/bin/activate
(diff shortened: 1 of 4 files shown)


.jenkins/actions/lint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e -x
+env_name=venv-${BUILD_NUMBER:-0}
+python3 -m venv ${env_name}
+. ${env_name}/bin/activate
+pip install -r fv3core/requirements/requirements_lint.txt -c constraints.txt
```
mcgibbon, line 6: We really should hard-code `pip install pre-commit` here since that's clearly the one command/package we're using for this action.

mcgibbon (approved):
LGTM, let's get those tests passing again!

PR #25, opened 2021-11-19
Feature/add shared

elynnwu:
- Add `shared` at the top level as an install package
    - package structure: pace.stencils
- Move `CubedToLatLon` to `shared`
- Keep `moist_cv` in `fv3core` for now because only 1 function is shared
- `shared` symbol link in `fv3core` the same way as `fv3gfs-util`


elynnwu:
launch jenkins

docker/Makefile
```diff
@@ -32,7 +32,7 @@ get_gt4py:
 
  ## build production container image
 fv3gfs_image: get_gt4py
-	if [ $(PULL) == True ]; then \
+	if [ $(PULL) = True ]; then \
```
elynnwu, line 35: Not sure why this is `=` after merging main, will fix this

.jenkins/install_virtualenv.sh
```diff
@@ -26,4 +26,5 @@ python3 -m pip install ${pace_dir}/fv3gfs-util/
 python3 -m pip install $wheel_command -c ${pace_dir}/constraints.txt -r fv3core/requirements/requirements_daint.txt
 python3 -m pip install ${PACE_INSTALL_FLAGS} ${pace_dir}/fv3core/
 python3 -m pip install ${PACE_INSTALL_FLAGS} ${pace_dir}/fv3gfs-physics/
+python3 -m pip install ${PACE_INSTALL_FLAGS} ${pace_dir}/shared/
```
elynnwu, line 29: We can also just have `pace/stencils` at the top level instead of `shared/pace/stencils`

shared/setup.cfg
```diff
@@ -0,0 +1,21 @@
...
+line_length = 88
+force_grid_wrap = 0
+include_trailing_comma = true
+multi_line_output = 3
+use_parentheses = true
+lines_after_imports = 2
+default_section = THIRDPARTY
+sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+known_first_party = fv3core,fv3gfs,pace
+known_third_party = f90nml,pytest,xarray,numpy,mpi4py,gt4py
```
mcgibbon, line 21: Because isort and flake8 are run by the top-level pre-commit, you shouldn't need to copy its configuration between subprojects. If I'm wrong and this breaks the tool, please let me know.

```suggestion
[aliases]

```

fv3gfs-physics/setup.cfg
```diff
@@ -17,5 +17,5 @@ use_parentheses = true
 lines_after_imports = 2
 default_section = THIRDPARTY
 sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
-known_first_party = fv3core,fv3gfs
+known_first_party = fv3core,fv3gfs,pace
```
mcgibbon, line 20: You should be able to delete this `setup.cfg` section as well.

dependencies.dot
```diff
@@ -21,5 +21,7 @@ digraph {
   fv3core -> util
   physics -> util
   physics -> fv3core
+  fv3core -> shared
+  physics -> shared
```
mcgibbon, line 25: It may be confusing to have "shared" show up in this graph when the module is named "stencils".

mcgibbon (commented):
We can discuss what to do on this, but as it stands `shared` has no testing infrastructure, and only gets tested as a part of other packages. We may decide to leave it that way for this PR. If there is a reason this should be left to later, like there's some refactoring we should do first to make it easier to do so, then I'd like to at least come up with what that task is and plan for it before merging.

We should also maybe talk about the "shared" name - what distinguishes this package from "utils", and why is the top level name "shared" different from the package name "stencils"? I'm not opposed to having multiple shared sub-modules under under one package, e.g. maybe we'd want to have "shared_gt4py" contain "fv3core.stencils" and "fv3core.gt4py" to keep these files out of "util". But we should try to name this package something that describes what it is in contrast to "util", because it will be harder to rename later.

fv3core/external/shared
```diff
@@ -0,0 +1 @@
+../../shared
```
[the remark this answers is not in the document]
elynnwu: Yes, good catch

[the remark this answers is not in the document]
elynnwu: Since physics does not have a docker file now, just adding the `stencils` mount in the makefile. In the future, we would want to have the same external that `fv3core` has.

dependencies.dot
elynnwu: Changed to `stencils`

elynnwu:
launch jenkins

elynnwu:
launch jenkins

elynnwu:
launch jenkins

elynnwu:
launch jenkins

dependencies.dot
```diff
@@ -9,17 +9,23 @@ digraph {
...
   physics [shape=oval, label="fv3gfs-physics"]
+  stencils [shape=oval, label="stencils"]
   util [shape=oval, label="fv3gfs-util"]
 
+
   pace -> fv3core
   pace -> physics
   pace -> util
   pace -> driver
+  pace -> stencils
```
mcgibbon, line 20: I expect pace will only indirectly depend on stencils - we probably won't instantiate its objects directly in a top-level example. However it's kind of odd in general that pace is in this file when it's not a subpackage, so this is fine either way.

mcgibbon (approved):
LGTM! One very minor comment.

PR #27, opened 2021-11-19
Add file caching infrastructure and test for fv3gfs-util using it

mcgibbon:
This PR adds the following:
- a script `.jenkins/cache.sh` which takes in a command "restore" or "save", a caching key, and a list of paths. "restore" will restore cached files, while "save" will save files the first time it is called, or print a message saying the cache already exists. This mirrors the behavior of CircleCI's caching mechanism.
-`fv3gfs-util/.jenkins/test_util.sh` which replaces the functionality of all other files in `fv3gfs-util/.jenkins`, making use of the new caching infrastructure. The other files can be removed once Jenkins is configured to use the new test script.


.jenkins/cache.sh
```diff
@@ -0,0 +1,49 @@
...
+key_cache=$cache_dir/$key
+target_dir=$(pwd)
+
+
+if [ "$cmd" == "save" ]; then
+    if [ -d $key_cache ]; then
+        echo "cache for key $key_cache already exists, skipping cache step"
+    else
+        mkdir -p $key_cache
+        for path in "${paths[@]}"; do
```
twicki, line 32: should there be a warning here if the remainder of paths is of size 0?

.jenkins/cache.sh
```diff
@@ -0,0 +1,49 @@
...
+            cp -r $path $key_cache/$path
+        done
+        echo "cache stored for key $key"
+    fi
+elif [ "$cmd" == "restore" ]; then
+    if [ -d $key_cache ]; then
+        cd $key_cache
+        for path in $(find . -maxdepth 1 -mindepth 1 -type d) ; do
+            echo $path
```
twicki, line 41: should this stay around?

.jenkins/cache.sh
```diff
@@ -0,0 +1,49 @@
...
+            mkdir -p $target_dir/$path
+            cp -r $path/* $target_dir/$path/
+        done
```
twicki, line 44: should there be a warning here one of the requested directories does not exist?

fv3gfs-util/.jenkins/test_util.sh
```diff
@@ -0,0 +1,47 @@
...
+    set +e
+    module unload cray-python
+    module unload pycuda
+    set -e
+    pip3 install -r $UTIL_DIR/requirements.txt -r $UTIL_DIR/requirements_gpu.txt -c $UTIL_DIR/constraints.txt -e $UTIL_DIR
+else
+    pip3 install -r $UTIL_DIR/requirements.txt -c $UTIL_DIR/constraints.txt -e $UTIL_DIR
+fi
+
+# pytest --junitxml results.xml $UTIL_DIR/tests
```
twicki, line 39: should this come back?

fv3gfs-util/.jenkins/test_util.sh
```diff
@@ -0,0 +1,47 @@
...
+. ./venv/bin/activate
+
+if [ "${target}" == "gpu" ] ; then
+    set +e
+    module unload cray-python
+    module unload pycuda
+    set -e
+    pip3 install -r $UTIL_DIR/requirements.txt -r $UTIL_DIR/requirements_gpu.txt -c $UTIL_DIR/constraints.txt -e $UTIL_DIR
+else
+    pip3 install -r $UTIL_DIR/requirements.txt -c $UTIL_DIR/constraints.txt -e $UTIL_DIR
```
twicki, line 36: why do we still need the pip installs if we copy in the venv?

.jenkins/checksum.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+echo $(md5sum $@ | md5sum | awk '{print $1;}')
```
twicki, line 5: what was the reasoning to exposing this on a higher level as compared to hiding it in `cache.sh`? I like that we have very fine grained control over what exactly is required to build the hash. Do we not consider this an implementation detail though?

twicki (changes requested):

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Test outputs
+results.xml
```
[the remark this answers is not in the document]
mcgibbon: Probably not, but this file affects not committing a file into git (which we reeeally don't want to do with this file), and not what files you do or don't delete after they're generated. There's also a nice git command to clear out all ignored files.

.jenkins/cache.sh
```diff
@@ -0,0 +1,49 @@
...
+unset paths[1]  # key is not a path to be cached
+
+if [ "$cmd" != "save" ] && [ "$cmd" != "restore" ]; then
+    echo "cmd must be one of 'save' or 'restore', got $cmd"
+    exit 1
+fi
+
+if [ -z $PACE_CACHE_DIR ]; then
+    cache_dir=~/.cache/pace
+    mkdir -p $cache_dir
```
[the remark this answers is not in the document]
mcgibbon: Good idea, I'll change it to create regardless.

.jenkins/cache.sh
mcgibbon: There should probably be an error. I'll update it to do so.

.jenkins/cache.sh
mcgibbon: no, will delete.

fv3gfs-util/.jenkins/test_util.sh
mcgibbon: Because the first time you run the script, the venv is empty. The way it is supposed to work is, this script should run identically if you delete the cache calls, but adding the cache lines makes it run faster while behaving the same.

fv3gfs-util/.jenkins/test_util.sh
mcgibbon: Yes. Very important. Will uncomment.

.jenkins/checksum.sh
mcgibbon: The short answer is because this is how CircleCI designed it.

The longer answer is, the files we are hashing as the cache key are entirely different than the files being cached, so we would need to add those as arguments to cache.sh. This isolates the tasks of defining a cache key and actually doing the caching into two scripts. The cache key can be anything - it doesn't have to be a file hash. Here we also add things like `v1` and whether this is for GPU or CPU.

.gitignore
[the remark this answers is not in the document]
twicki: I'm happy with that still in here. This was mostly curiousity.

.jenkins/cache.sh
mcgibbon: I don't think so. These paths are not necessarily expected to exist when the cache is unpacked. On the first execution (without a cache), they can be created further down the line before the cache is saved.

fv3gfs-util/.jenkins/test_util.sh
mcgibbon: Uncommented.

mcgibbon (commented):
Should be ready for re-review.

.jenkins/cache.sh
twicki: ok

fv3gfs-util/.jenkins/test_util.sh
twicki: I see.
The only question I have is do we really gain much from copying the venv then?
Checking if all the requirements are fulfilled is about as fast as getting the wheels from somewhere, right? And either way we need a persistent storage somewhere. 
I'm not expecting an answer in absolute time, but I'm just curious if you have a gut feeling?

.jenkins/checksum.sh
twicki: that makes sense

fv3gfs-util/.jenkins/test_util.sh
mcgibbon: > Checking if all the requirements are fulfilled is about as fast as getting the wheels from somewhere, right?

This is much faster in my experience, even when all the dependencies are wheels. But also it is very hard to ensure all the dependencies are wheels. Our current pace-fv3core-daint-setup plan last took 4 minutes 22 seconds to run, and the only thing it does other than install python packages appears to be linting.

twicki (approved):
lgtm

mcgibbon:
launch jenkins

PR #31, opened 2021-11-23
baroclinic initialization

rheacangeo:
### Purpose
To add the computation of the baroclinic perturbation test case (Jablonowski & Williamson JRMS 2006) to enable running performance scripts without reading serialized data. 

This PR:
*  adds validating code (against data generated with the feature/serialize_init branch of fv3gfs-fortran, which will in parallel be PRed into master as data version 7.2.7) that creates the baroclinic initial state in fv3core/fv3core/initialization
* Two files distinguish the test case -- baroclinic_jablonowski_williamson.py has the base equations found in the paper and DCMIP2016 operating on numpy arrays and baroclinic.py, which calls the functions from the other file, and does the cubed sphere transformation and data shaping sizing to get it in the format we can use in the Dycore. 
* Adds code to test the initial state values when run with baroclinic namelists (not currently blocking PRs, but does or will when it works, get tested with the cache generating cron plan). 
* adds a DycoreState specification using a dataclass and replaces the ArgSpec and get_namespace functionality of fv3core/fv3ocre/stencils/fv_dynamics.py
* Updated the fv_dynamics test to use the DycoreState
* updates the dynamics runfile for performance testing to create a DycoreState using the baroclinic initialization and replaced reading the serialized data as inputs.


fv3core/examples/standalone/runfile/dynamics.py
```diff
@@ -215,20 +219,43 @@ def collect_data_and_write_to_file(
...
+            adiabatic=namelist.adiabatic,
+            hydrostatic=namelist.hydrostatic,
+            moist_phys=namelist.moist_phys,
+            comm=communicator,
+        )
+
         dycore = fv3core.DynamicalCore(
             comm=communicator,
-            grid_data=grid.grid_data,
+            grid_data=GridData.new_from_metric_terms(metric_terms),
```
mcgibbon, line 247: Nice to see this decoupled from Grid!

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+        jablo_init.surface_geopotential_perturbation,
+        [],
+        lat,
+        lat_agrid,
+        lat2,
+        lat3,
+        lat4,
+        lat5,
+        slice_2d,
+    )
```
mcgibbon, line 325: `cell_average_nine_components` and `compute_grid_edge_midpoint_latitude_components` are doing something low level and pretty complex - I can't tell what they are doing. compute_grid_edge_midpoint_latitude_components is also only used for this second function.

Is there any change you can make to have it be more clear what's happening at this function's level, by pushing the low-level logic down one call level? For example, even if it's less efficient, by pushing the call to compute_grid_edge_midpoint_latitude_components down one level (removing many low-level call arguments from this function's level), and by pulling the slicing operation (which is a higher level operation you're already using here in the assignment slice) up to this level (e.g. by passing `lat[slice_2d]` and `lat_agrid[slice_2d]`)?

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+):
+    """
+    Computes auxiliary pressure variables for a hydrostatic state.
+
+    The Fortran code also recomputes some more pressure variables,
+    pe, pk, but since these are already done in setup_pressure_fields
+    we don't duplicate them here
+    """
+    assert not adjust_dry_mass
+    assert not hydrostatic
```
mcgibbon, line 390: Does it make sense for us to just not implement these as keyword arguments? I don't think we have any plans to implement these, the code itself is easier to understand than the rest of the dycore in terms of if we were to implement a new option, and if we did implement a new option we'd probably do it using new functions rather than if flags?

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+        pe=state.pe.data[slice_3d],
+        peln=state.peln.data[slice_3d],
+        pkz=state.pkz.data[slice_3d],
+        ptop=metric_terms.ptop,
+        moist_phys=moist_phys,
+        make_nh=(not hydrostatic),
+        hydrostatic=hydrostatic,
+    )
+    # TODO: when the dycore state is updated to only include
+    # quantities and no storages, remove the "_quantity" from phis, u and v
```
mcgibbon, line 517: You could remove this TODO here, when I go to do it this will be using vscode's refactor functionality so it will update everywhere at once (thanks to your use of a dataclass!). As it stands, this comment would probably remain after the refactor unless a reviewer catches it or I've thought long enough about this particular comment location while writing this message.
```suggestion
```

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+        pk=state.pk.data[slice_3d],
+        pkz=state.pkz.data[slice_3d],
+        ak=metric_terms.ak.data,
+        bk=metric_terms.bk.data,
+        ptop=metric_terms.ptop,
+        lat_agrid=metric_terms.lat_agrid.data[slice_2d],
+        adiabatic=adiabatic,
+    )
+    baroclinic_initialization(
+        eta=eta,
```
mcgibbon, line 476: I appreciate the use of keyword arguments!

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+    state.w.data[:] = 1.0e30
+    state.delz.data[:] = 1.0e25
+    state.phis.data[:] = 1.0e25
+    state.ps.data[:] = jablo_init.surface_pressure
+    eta = np.zeros(nz)
+    eta_v = np.zeros(nz)
+    islice, jslice, slice_3d, slice_2d = compute_slices(nx, ny)
+    # Slices with extra buffer points in the horizontal dimension
+    # to accomodate averaging over shifted calculations on the grid
+    isliceb, jsliceb, slice_3db, slice_2db = compute_slices(nx + 1, ny + 1)
```
mcgibbon, line 455: I get that b means b-grid which should mean interfaces (or maybe it means buffer?), but this is one of the harder to understand naming conventions for grad students - is there any way to avoid this?

Two of the outputs appear to be unused:
```suggestion
    _, _, slice_3db, slice_2db = compute_slices(nx + 1, ny + 1)
```

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,522 @@
...
+
+    # TODO: It would be neat to use <quantity>.view here rather than slices
+    # But doesn't work with all numpy operations, would need to either
+    # support some features or change the calculations
```
mcgibbon, line 459: Yeah, the `view` interface was really hard to come up with any design for and the one in place is kind of confusing. Maybe `quantity.view(origin_add=tuple, domain_add=tuple)`?

fv3core/fv3core/initialization/baroclinic_jablonowski_williamson.py
```diff
@@ -0,0 +1,166 @@
...
+pt0 = 0.0
+eta_0 = 0.252
+eta_s = 1.0  # surface level
+eta_t = 0.2  # tropopause
+t_0 = 288.0
+delta_t = 480000.0
+lapse_rate = 0.005  # From Table VI of DCMIP2016
+surface_pressure = 1.0e5  # units of (Pa), from Table VI of DCMIP2016
+# NOTE RADIUS = 6.3712e6 in FV3 vs Jabowski paper 6.371229e6
+R = constants.RADIUS / 10.0  # Perturbation radiusfor test case 13
```
mcgibbon, line 30: Good comments! Here and throughout the PR.

fv3core/fv3core/initialization/baroclinic_jablonowski_williamson.py
```diff
@@ -0,0 +1,166 @@
...
+# maximum windspeed amplitude - close to windspeed of zonal-mean time-mean
+# jet stream in troposphere
+u0 = 35.0  # From Table VI of DCMIP2016
+# [lon, lat] of zonal wind perturbation centerpoint at 20E, 40N
+pcen = [math.pi / 9.0, 2.0 * math.pi / 9.0]  # From Table VI of DCMIP2016
+u1 = 1.0
+pt0 = 0.0
+eta_0 = 0.252
+eta_s = 1.0  # surface level
+eta_t = 0.2  # tropopause
```
mcgibbon, line 24: These are each only used in one place - can they be renamed to `eta_surface` and `eta_tropopause`?

fv3core/fv3core/initialization/dycore_state.py
```diff
@@ -0,0 +1,380 @@
...
+        for field_name in field_names:
+            if field_name not in dict_of_numpy_arrays.keys():
+                raise KeyError(
+                    field_name
+                    + " is not included in the provided dictionary of numpy arrays"
+                )
+        return state
+
+    @classmethod
+    def init_from_quantities(cls, dict_of_quantities):
```
mcgibbon, line 329: Thanks for this feature!

fv3core/fv3core/initialization/dycore_state.py
```diff
@@ -0,0 +1,380 @@
...
+        translate_object = fv3core.testing.TranslateFVDynamics([grid])
+        input_data = translate_object.collect_input_data(serializer, savepoint_in)
+        # making just storages for the moment, revisit when making them all
+        # quantities (maybe use state_from_inputs)
+        translate_object._base.make_storage_data_input_vars(input_data)
+        # used for the translate test as inputs, but are generated by the
+        # MetricsTerms class and are not part of this data class
+        for delvar in ["ak", "bk", "ptop", "ks"]:
+            del input_data[delvar]
+        return cls(**input_data, quantity_factory=quantity_factory)
```
mcgibbon, line 377: This method should only be getting used by the tests, and currently fv3core doesn't depend on serialbox other than for testing. Can this get moved to the testing code?

fv3core/fv3core/initialization/dycore_state.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def __getitem__(self, item):
```
mcgibbon, line 379: This is a good feature for minimal changes for this PR, but we should plan to refactor getitem calls because this removes type safety/inspection checking. For example, when I go to use vscode's refactor on attribute names for this class, it will be able to automatically refactor all `state.item` accesses, but not the `state["item"]` accesses. I will probably do this as an earlier commit in that PR.

fv3core/fv3core/testing/translate.py
```diff
@@ -293,6 +296,12 @@ def make_grid_storage(self, pygrid):
             if k in self.data:
                 self.make_composite_var_storage(k, self.data[k], shape)
                 del self.data[k]
+        for k in TranslateGrid.ee_vars:
```
mcgibbon, line 299: nit: This was pre-existing, but can you replace `k` here and below with `key`? When I read these my mind doesn't stop telling me it's indexing on the vertical direction.

fv3core/fv3core/testing/translate_fvdynamics.py
```diff
@@ -282,15 +284,27 @@ def compute_parallel(self, inputs, communicator):
             grid_data.ptop = inputs["ptop"]
             grid_data.ks = inputs["ks"]
 
-        state = self.state_from_inputs(inputs)
+        input_storages = self.state_from_inputs(inputs)
+        # making sure we init DyCOreState with the exact set of variables
```
mcgibbon, line 288:
```suggestion
        # making sure we init DyCoreState with the exact set of variables
```

fv3gfs-util/fv3gfs/util/quantity.py
```diff
@@ -183,7 +183,7 @@ def _get_compute_index(self, index):
         if len(index) > len(self._dims):
             raise IndexError(
                 f"{len(index)} is too many indices for a "
-                f"{len(self.dims)}-dimensional quantity"
+                f"{len(self._dims)}-dimensional quantity"
```
mcgibbon, line 186: nice catch!

fv3core/fv3core/initialization/baroclinic.py
rheacangeo: ok thanks

fv3core/fv3core/initialization/baroclinic_jablonowski_williamson.py
rheacangeo: sure thing

fv3core/fv3core/initialization/baroclinic_jablonowski_williamson.py
rheacangeo: thanks! paid close attention to the paper on this one, which thankfully matches pretty well to what the fortran code does!

fv3core/fv3core/initialization/dycore_state.py
rheacangeo: It can. I was thinking it might be useful for performance testing if we end up wanting to test non-baroclinic namelists before we implement reading initial state from disk. But someone wanting to do that can probably figure it out.

fv3core/fv3core/initialization/dycore_state.py
rheacangeo: Ok thanks for the heads up.

fv3core/fv3core/testing/translate.py
rheacangeo: sure thing

rheacangeo (commented):
Thanks for the quick turnaround on feedback!

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,506 @@
...
+    )
+
+
+def cell_average_nine_components(
+    component_function,
+    component_args,
+    lon,
+    lat,
+    lat_agrid,
+    grid_slice,
```
mcgibbon, line 180: The type of this argument is non-intuitive and would be particularly helpful to type hint.
```suggestion
    grid_slice: Tuple[slice, slice],
```

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -0,0 +1,506 @@
...
+):
+    """
+    Componet 9 cell components of a variable, calling a component_function
+    definiting the equation with component_args (arguments unique to that
+    component_function), and precomputed lat arrays
+    """
```
mcgibbon, line 186:
```suggestion
    """
    Outputs the weighted average of a field that is a function of latitude,
    averaging over the 9 points on the corners, edges, and center of each
    gridcell.
    
    Args:
        component_function: callable taking in an array of latitude and
            returning an output array
        component_args: arguments to pass on to component_function,
            should not be a function of latitude
        lon: longitude array, defined on cell corners
        lat: latitude array, defined on cell corners
        lat_agrid: latitude array, defined on cell centers
        grid_slice: tuple of x-domain and y-domain slices defining the compute domain
            for cell-centered variables
    """
    # this weighting is done to reproduce the behavior of the Fortran code
```

fv3core/fv3core/initialization/baroclinic.py
mcgibbon: If you do still want to pull `grid_slice` up one level, the way you would do it is by passing in `lon` and `lat` sliced on their own compute domains - that is, up to nx+1 and ny+1 since they're defined on cell corners. Then below, every access to lat or lon should be clipping off 1 point. pt6 becomes `lat[:-1, :-1]` while pt9 becomes `lat[:-1, 1:]`.

mcgibbon (approved):
Looks great! Looking forward to making use of this.

PR #33, opened 2021-11-24
Feature/shrinked edge partitioning

yniederm:
- Add `edge_interior_ratio` as an initialization argument for `TilePartitioner`.
- This can later be used in `TilePartitioner.subtile_slice()` to do custom subdomain partitioning.
This is based on the idea that ranks operating on boundary subdomains should have a smaller domain assigned. 
- `tests/test_partitioner.py` updated to reflect some more advanced assignments.


ofuhrer:
Can one of the admins verify this patch?

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -812,17 +843,33 @@ def subtile_slice(
...
             end = start + index.extent
         else:
             end = start + index.base_extent
-        return_list.append(slice(start, end))
+        return_list.append(slice(int(start), int(end)))
     return tuple(return_list)
+
+
+def normal_round(n):
+    "Needed to get around IEEE754 half-to-even rounding. Will not work with negative numbers."
```
mcgibbon, line 872: Why is it bad to have half-to-even rounding here? As a user, I'd probably expect rounding to occur with the behavior of Python's default `round`.

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -108,150 +107,436 @@ def test_tile_extent_from_rank_metadata(array_extent, array_dims, layout, tile_e
...
 @pytest.mark.cpu_only
 def test_subtile_slice(
-    array_dims, tile_extent, layout, subtile_index, subtile_slice, overlap
+    array_dims, tile_extent, layout,
+    rank, subtile_slice, overlap, edge_tile_ratio
 ):
-    result = fv3gfs.util.partitioner.subtile_slice(
-        array_dims, tile_extent, layout, subtile_index, overlap
+    partitioner = fv3gfs.util.TilePartitioner(layout, edge_tile_ratio)
+    result = partitioner.subtile_slice(
```
mcgibbon, line 539: nit: in this refactor, could you put `rank` as the third argument so it's in the same position as the two arguments it's replacing (layout and subtile_index)?

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -108,150 +107,436 @@ def test_tile_extent_from_rank_metadata(array_extent, array_dims, layout, tile_e
...
+            (5, 16, 16),
+            (4, 4),
+            14,
+            (slice(0, 5), slice(14, 16), slice(8, 14)),
+            False,
+            float(1 / 3),
+            id="96_rank_fartopcloseright_right_no_overlap_square_layout_third_edge_tiles_3d",
+        ),
+        pytest.param(
+            [fv3gfs.util.Y_INTERFACE_DIM, fv3gfs.util.X_INTERFACE_DIM],
```
mcgibbon, line 492: The case we usually see for interfaces is that one dimension is an interface dimension and the other is not - could you add one or two cases like this?

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -108,150 +107,436 @@ def test_tile_extent_from_rank_metadata(array_extent, array_dims, layout, tile_e
...
+            id="54_rank_topright_right_no_overlap_square_layout_half_edge_tiles_3d",
+        ),
+        pytest.param(
+            [fv3gfs.util.Z_INTERFACE_DIM, fv3gfs.util.Y_DIM, fv3gfs.util.X_DIM],
+            (5, 8, 8),
+            (3, 3),
+            0,
+            (slice(0, 5), slice(0, 1), slice(0, 1)),
+            False,
+            float(1 / 6),
```
mcgibbon, line 408: FYI: Integer division always returns a float in python 3, but you can be really sure by using a dot to make one of the numbers a float like in `1./6`.

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -108,150 +107,436 @@ def test_tile_extent_from_rank_metadata(array_extent, array_dims, layout, tile_e
...
+        pytest.param(
+            [fv3gfs.util.Y_DIM, fv3gfs.util.X_DIM],
+            (12, 12),
+            (2, 4),
+            0,
+            (slice(0, 6), slice(0, 2)),
+            True,
+            0.5,
+            id="48_rank_botfarleft_right_overlap_half_edge",
+        ),
```
mcgibbon, line 530: If you can, please add a test case or an entirely separate test for layout=(5, 5), array size = (16, 16), dims=[X_DIM, Y_DIM], tile_ratio=1. The "correct" decomposition from Fortran has length per rank of (3, 3, 4, 3, 3). The current code does not support this case - you can make a test that simply asserts an exception should be raised (a ValueError) by using
```
with pytest.raises(ValidationError):
    function_that_raises_a_validation_error()
```
That line will fail if the expected exception is not raised and continue if it is raised.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -723,24 +727,49 @@ def tile_extent_from_rank_metadata(
 
 
 def rank_extent_from_tile_metadata(
-    dims: Sequence[str], tile_extent: Sequence[int], layout: Tuple[int, int]
-) -> Tuple[int, ...]:
+        dims: Sequence[str], tile_extent: Sequence[int], subtile_index: Tuple[int, int],
```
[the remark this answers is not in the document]
mcgibbon: If all the current tests pass without this change, then it means scatter/gather only breaks when `edge_tile_ratio` on the TilePartitioner is not 1. That would mean we can consider merging this code first, then adding a test combining these two features (which would fail) and fixing the code in another PR.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -723,24 +727,49 @@ def tile_extent_from_rank_metadata(
...
 
     Args:
         dims: dimension names
-        rank_extent: the extent of a tile
+        tile_extent: the extent of a tile
+        subtile_index: the (y, x) position of the rank on the tile
         layout: the (y, x) number of ranks along each tile axis
+        edge_tile_ratio (optional): ratio between interior and boundary tile sizes.
+            Assumes full integer divisibility for both interior 
+            and boundary tile sizes.
```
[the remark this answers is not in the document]
yniederm: You're right, `edge_interior_ratio` makes more sense. Let's say along an axis you have a rank that's located on the edge, which is half as large as the interior subdomains, then an `edge_interior_ratio` of 1:2 = 0.5 seems like accurate nomenclature.

fv3gfs-util/tests/test_partitioner.py
yniederm: Done.

fv3gfs-util/fv3gfs/util/partitioner.py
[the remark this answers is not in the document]
mcgibbon: Since you changed the call signature to this routine, could you add new test cases that make use of the new arguments and refactor the existing tests to pass (i.e. provide these new required arguments)?

fv3gfs-util/tests/test_partitioner.py
yniederm: Done

twicki:
ok to test

mcgibbon:
launch jenkins

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Behaviour changed to use native python rounding. I found avoiding half-to-even-rounding gave better results, but with the new approach this doesn't matter anymore.

fv3gfs-util/tests/test_partitioner.py
yniederm: Rerunning the `pytest` tests, it seems a _lot_ of calls to subtile_slice are used and all of them have the rank as first argument, should I change all the tests to use rank as thirdargument of leave it as is (first argument)?

Here's one example:
`tests/test_tile_scatter_gather.py::test_tile_gather_no_recv_quantity[y,z,x-3-layout0-0-numpy] - TypeError: subtile_slice() got multiple values for argument 'global_dims'`
with the corresponding call:
`subtile_slice = partitioner.subtile_slice(
            rank,
            global_dims=tile_quantity.dims,
            global_extent=tile_quantity.extent,
            overlap=True,
        )`

fv3gfs-util/tests/test_partitioner.py
yniederm: Test cases have been added for both variants (just X or Y interfaces).

fv3gfs-util/tests/test_partitioner.py
yniederm: The code now supports this case, it will find an valid assignment of (2,4,4,4,2). 
I have however added tests for a case where the grid has an odd number of points along an axis and the number of ranks along the same axis is even.

fv3gfs-util/fv3gfs/util/partitioner.py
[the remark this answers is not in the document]
yniederm: Done. Scatter/Gather seem to work with an unchanged `edge_interior_ratio` as all the current tests are passing.

yniederm:
This PR is replaced by #44, more information can be found in #44.

PR #39, opened 2021-12-01
Revert "Restore buildenv as a submodule instead of hard files"

mcgibbon:
Reverts ai2cm/pace#38, due to submodule being placed in the wrong location.


elynnwu (approved):

mcgibbon:
Turns out not to be necessary since nothing is referencing external/buildenv, and in the end we'd like it to be placed at the top level to be managed by the `changed_from_main.py` dependency tree.

PR #40, opened 2021-12-02
Add required backend arg to make_storage methods

mcgibbon:
Currently, make_storage helper methods use the globally-set backend. This PR updates these methods to explicitly take in the backend used for the storage.

Changes:
- backend is added as a required keyword argument to `dyncore_temporaries`, `TranslateGrid`, `storage_dict`, `make_grid`, `process_grid_savepoint` and functions that call it in conftest.py, `make_storage_from_shape`, `make_storage_from_shape_uncached`, `make_storage_data`, `get_quantity_halo_spec`, `get_halo_update_spec`, and `make_storage_dict`
- Type hints are added to Quantity attributes on MetricTerms
- .dace.conf is added to .gitignore


mcgibbon:
launch jenkins

mcgibbon:
Tests passed apart from 2 failures on 54 ranks that appear to be due to distributed compilation (the "<name> has no attribute <name>" error where <name> is an auto-generated stencil module).

fv3core/examples/standalone/runfile/acoustics.py
```diff
@@ -39,20 +39,6 @@ def initialize_serializer(data_directory: str, rank: int = 0) -> serialbox.Seria
...
-    """Uses the serializer to generate a Grid object from serialized data"""
-    grid_savepoint = serializer.get_savepoint("Grid-Info")[0]
-    grid_data = {}
-    grid_fields = serializer.fields_at_savepoint(grid_savepoint)
-    for field in grid_fields:
-        grid_data[field] = serializer.read(field, grid_savepoint)
-        if len(grid_data[field].flatten()) == 1:
-            grid_data[field] = grid_data[field][0]
-    return fv3core.testing.TranslateGrid(grid_data, rank).python_grid()
-
```
rheacangeo, line 54: actually we'd prefer to not delete this yet -- it still might be used for running non-baroclinic configurations.

fv3core/examples/wrapped/runfiles/baroclinic_test.py
```diff
@@ -124,7 +124,8 @@ def convert_3d_to_1d(state, field_names):
     dt_atmos = spec.namelist.dt_atmos
 
     # set backend
-    fv3core.set_backend("numpy")
+    backend = "numpy"
+    fv3core.set_backend(backend)
```
rheacangeo, line 128: so we are leaving the 'global_config' global state? does this still need to be set?

fv3core/fv3core/stencils/a2b_ord4.py
```diff
@@ -489,9 +489,14 @@ def __init__(
 
         self.replace = replace
 
-        self._tmp_qx = utils.make_storage_from_shape(self._idx.max_shape)
-        self._tmp_qy = utils.make_storage_from_shape(self._idx.max_shape)
-        self._tmp_qout_edges = utils.make_storage_from_shape(self._idx.max_shape)
+        def make_storage():
+            return utils.make_storage_from_shape(
+                self._idx.max_shape, backend=self._stencil_config.backend
+            )
```
rheacangeo, line 495: yay!

elynnwu (commented):
Wow! That's a lot of `make_storage`. Thanks for going through this!

fv3core/fv3core/stencils/c_sw.py
```diff
@@ -416,22 +423,31 @@ def __init__(
...
+        def make_storage():
+            return utils.make_storage_from_shape(
+                grid_indexing.max_shape, backend=stencil_factory.backend
+            )
+
+        self._tmp_ke = make_storage()
+        self._tmp_vort = make_storage()
+        self._tmp_fx = make_storage()
+        self._tmp_fx1 = make_storage()
+        self._tmp_fx2 = make_storage()
```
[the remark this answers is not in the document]
elynnwu: Curious to see what others think...

fv3core/fv3core/stencils/a2b_ord4.py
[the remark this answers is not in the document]
mcgibbon: The alternative design on this would be to have a `make_storage` method on the `StencilConfig` and `StencilFactory` classes, which makes use of its backend. @rheacangeo and I settled on this design so that you can still make storages without the larger configuration needs of StencilConfig, but it's possible we could enable both. Then here we could have `stencil_factory.make_storage(self._idx.max_shape)`.

Perhaps something to discuss in a team meeting.

fv3core/fv3core/stencils/c_sw.py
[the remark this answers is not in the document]
mcgibbon: I don't know if there's one standard that's best for every module - it depends a lot on how many allocations there are, how many of them use exactly the same arguments (as these do), how many use different args and which ones are different, whether `origin` is being set, whether `backend` already exists as its own variable in the init, how long the variable names being set are, etc. etc. . In some modules, I did go the explicit argument route, but when it would multiply a large number of allocation lines by 3, I opted for this pattern.

The balance would also be different if e.g. we renamed make_storage_from_shape to make_storage.

fv3core/examples/standalone/runfile/acoustics.py
mcgibbon: Would you say after these tests running a few days that it's stable enough to remove this code from the file, and use the git history to recover it if we need it? If not I can put it back, in which case when should we re-assess this?

fv3core/examples/wrapped/runfiles/baroclinic_test.py
mcgibbon: Yes, it's going to be a bigger task to fully remove the global config. After this PR I'll go back to my PR running a timestep of the dynamics with no global state access, and once that test is done I can copy the pattern used to these runfiles. After that, we'll still need to refactor the unit tests to pass `Grid` explicitly to the tests instead of moving through global state. After all that, we *should* be able to remove this.

mcgibbon:
launch jenkins

fv3core/examples/wrapped/runfiles/baroclinic_test.py
mcgibbon: In this specific case, we use `spec.grid` below, and `Grid` still makes use of global_config to get the backend.

fv3core/fv3core/stencils/a2b_ord4.py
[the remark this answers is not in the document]
elynnwu: That makes sense. I'll put this on the team meeting next week.

fv3core/examples/standalone/runfile/acoustics.py
rheacangeo: let's remove it yeah.

fv3core/examples/wrapped/runfiles/baroclinic_test.py
rheacangeo: thanks!

rheacangeo (approved):
lgtm, but also Elynn brings up good questions!

elynnwu (approved):

PR #41, opened 2021-12-02
run_on_daint fixed for more layouts

rheacangeo:
With the shift to computing grid variables and state on the fly rather than using serialized data, we had broken run_on_daint's ability to run other layouts besides 1, 1. This should fix this.


twicki (approved):
lgtm

rheacangeo:
launch jenkins

PR #42, opened 2021-12-02
Enable flake8 tests on fv3gfs-util

mcgibbon:


.pre-commit-config.yaml
```diff
@@ -50,7 +50,7 @@ repos:
     - id: trailing-whitespace
 
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
+    rev: 3.9.2
```
mcgibbon, line 53: This was updated due to a bug in an earlier version that causes flake8 to error on `# type: ignore[flag]` comments

FlorianDeconinck (approved):

PR #44, opened 2021-12-03
Add edge_interior_ratio argument to TilePartitioner

yniederm:
- `tests/test_partitioner.py` updated to reflect some more advanced assignments.
- `subtile_extent` method of Partitioner classes now takes in a required `rank` argument
- TilePartitioner has a new `edge_interior_ratio` argument which defaults to 1.0, and lets the user specify the relative 1-dimensional extent of edge and corner compute domains compared to interior compute domains. In all cases, the closest valid value will be used, which enables some previously invalid configurations (e.g. C128 on a 3 by 3 layout will use the closest valid edge_interior_ratio to 1.0)
- This can later be used in `TilePartitioner.subtile_slice()` to do custom subdomain partitioning.

This PR is replacing #33 with its feature branch moved from my fork to the `pace` repo.


ofuhrer:
Can one of the admins verify this patch?

mcgibbon:
ok to test

twicki:
launch jenkins

twicki:
launch jenkins

twicki:
launch jenkins

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -68,9 +68,9 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
     @abc.abstractmethod
     def subtile_slice(
         self,
-        rank: int,
         global_dims: Sequence[str],
         global_extent: Sequence[int],
+        rank: int,
```
mcgibbon, line 73: Please keep the same ordering of the arguments, in case the ML team is depending on this ordering.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -141,17 +143,38 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
             rank_metadata.dims, rank_metadata.extent, self.layout
         )
 
-    def subtile_extent(self, global_metadata: QuantityMetadata) -> Tuple[int, ...]:
+    def subtile_extent(self, global_metadata: QuantityMetadata, rank: int, overlap: bool, cube_metadata_offset=0) -> Tuple[int, ...]:
```
mcgibbon, line 146: What does "cube_metadata_offset" mean? Can you add a list of `Args:` to the docstring where you describe it?

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -141,17 +143,38 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
...
+            return_extent.append(_interface_overlap_extent(dim, is_end_index, tile_decomposition[num_dim + edge_index_offset], overlap))
+
+        return tuple(return_extent)
 
     def subtile_slice(
         self,
-        rank: int,
         global_dims: Sequence[str],
         global_extent: Sequence[int],
+        rank: int,
```
mcgibbon, line 177: Please keep call signatures the same where possible.

fv3gfs-util/fv3gfs/util/partitioner.py
mcgibbon: I don't think `overlap` should be an argument for this, right? It would always be True, since we want the size of the numpy array/storage on this rank.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -568,22 +592,21 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
             rank_metadata.dims, rank_metadata.extent, self.layout
         )
 
-    def subtile_extent(self, cube_metadata: QuantityMetadata) -> Tuple[int, ...]:
+    def subtile_extent(self, cube_metadata: QuantityMetadata, rank: int, overlap: bool) -> Tuple[int, ...]:
```
mcgibbon, line 595: I think overlap should always be `True` here. Subtile "extent" inherently should include overlapping points, it's when you're slicing a full tile and want to ensure subtile data doesn't overlap that you need the subtile "slice" not to include overlap.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -568,22 +592,21 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
...
-        )
+
+        return self.tile.subtile_extent(cube_metadata, rank, overlap, cube_metadata_offset=1)
 
     def subtile_slice(
         self,
-        rank: int,
         global_dims: Sequence[str],
         global_extent: Sequence[int],
+        rank: int,
```
mcgibbon, line 609: Please keep call signatures the same where possible.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -716,31 +738,95 @@ def tile_extent_from_rank_metadata(
     Returns:
         tile_extent: the extent of one tile
     """
+    if edge_interior_ratio < 1.0:
+        print(f"{edge_interior_ratio}", flush=True)
+        raise NotImplementedError("Only equal sized subdomains are supported.")
```
mcgibbon, line 743: Usually better to include debugging information in the exception, print statements shouldn't exist in the final merged code.

```suggestion
        raise NotImplementedError(
            "Only equal sized subdomains are supported, was given "
            f"an edge_interior_ratio of {edge_interior_ratio}"
        )
```

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -1,3 +1,5 @@
+import copy
```
mcgibbon, line 1: Unused import, should be picked up by pre-commit.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Changed all overlaps to be `True` by default.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: `cube_metadata_offset` was unhelpfully named. It's now a boolean `is_cube_metadata` that tells `subtile_extent` that the metadata that's passed in is that of a cubed Sphere model. It's used to offset the `dims` and `extent` members of `global_metadata`, like in this call: `global_metadata.dims[cube_metadata_offset:]`.
A docstring has been added.

yniederm: You're right, `overlap` should always be true in that function. Replaced by `True`.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Has been changed. Not sure how that snuck by.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Reverted the argument order change.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Reverted the argument order change.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Reverted the argument order change.

mcgibbon:
launch jenkins

fv3gfs-util/fv3gfs/util/partitioner.py
mcgibbon: Why is overlap an argument? When should it be `False`?

yniederm: Overlap as an argument mirrors `subtile_slice`, where it is also one. The way the code currently calculates the extents of a subtile incorporates `overlap` as an argument. This function uses the same mechanism, and it assumes that `overlap` _can_ be `False`.

mcgibbon: But when `subtile_slice` calls `subtile_extent`, it will always pass `overlap=True`. It will give an incorrect answer if it passes `overlap=False`. This is not a value that should be modified inside this function (unless you can describe a use case where it should be modified to False).

mcgibbon: Another way to put it is this - if you are going to keep this as an argument, what would you put in the docstring to describe what it does, and when you would want to set it to what value?

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -141,18 +158,79 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
...
+            cube_metadata_offset = 1
+        else:
+            cube_metadata_offset = 0
+
+        # cube_metadata_offset is used to exclude the TILE dimension (first index) for cubed sphere metadata.
+        tile_decomposition = subtile_extents_from_tile_metadata(
+            global_metadata.dims[cube_metadata_offset:],
+            global_metadata.extent[cube_metadata_offset:],
+            self.layout,
         )
```
mcgibbon, line 188: The tile dimension is not necessarily guaranteed to be the first dimension, but it is guaranteed to have the value `pace.util.TILE_DIM`. Can you remove the `is_cube_metadata` argument and detect the presence of (and remove) the TILE_DIM instead?

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -712,39 +809,136 @@ def tile_extent_from_rank_metadata(
...
+        )
     layout_factors = np.asarray(
         utils.list_by_dims(dims, layout, non_horizontal_value=1)
     )
     return extent_from_metadata(dims, rank_extent, layout_factors)
 
 
-def rank_extent_from_tile_metadata(
-    dims: Sequence[str], tile_extent: Sequence[int], layout: Tuple[int, int]
+def listArgsToTupleArgs(function):
```
mcgibbon, line 830: Please follow pothole_case for function names (including decorators).

Neat idea! I believe this will currently break if I call `subtile_extents_from_tile_metadata(dims=dims, tile_extent=tile_extent, layout=layout, edge_interior_ratio=ratio)`. Can you fix it to work with kwargs as well? You may want to write a unit test for it.

fv3gfs-util/tests/test_partitioner.py
```diff
@@ -108,150 +108,778 @@ def test_tile_extent_from_rank_metadata(array_extent, array_dims, layout, tile_e
...
+            0.5,
+            (slice(0, 3, None), slice(0, 2, None)),
+            (9, 8),
+            "Only equal sized subdomains are supported, was given an edge_interior_ratio of 0.5",
+            id="48_rank_half_edge_tiles",
+        ),
+    ],
+)
+@pytest.mark.cpu_only
+def test_tile_extent_from_metadata(
```
mcgibbon, line 861: Good job testing the other changes you made!

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -712,39 +809,136 @@ def tile_extent_from_rank_metadata(
...
+                raise ValueError(
+                    f"No valid subdomain assignment found for dimension {dim} with {dim_extent} gridpoints along {subtile_count} ranks."
+                )
+            return_extents.append(int(edge_size / dim_edge_interior_ratio))
+            edge_extents.append(int(edge_size))
+        else:
+            edge_size = int(dim_extent / subtile_count)
+            return_extents.append(edge_size)
+            edge_extents.append(edge_size)
+    return_extents.extend(edge_extents)
```
mcgibbon, line 936: This is pretty long and complex. In some cases I'd ask the code author to refactor this into smaller bits to be clearer. Instead can you add a description somewhere, either as a multi-line comment or as comments throughout the code, of the general algorithm you've implemented here for solving the problem?

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -812,17 +968,66 @@ def subtile_slice(
...
+                    dim, False, subtile_extent[num_dim], overlap
+                )
+            horizontal_dim_index = horizontal_dim_index + 1
         else:
-            end = start + index.base_extent
-        return_list.append(slice(start, end))
+            start = 0
+            end = dim_extent
+
+        return_list.append(slice(int(start), int(end)))
```
mcgibbon, line 1024: This code looks very similar to the other large, complex block of code. Are there any shared functionalities you can refactor out into a helper routine? It's hard for me to tell because I can't quite tell what the codes are doing.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: `subtile_slice` never calls `subtile_extent`, it just has an int tuple named like that. Which should probably be changed to `subtile_extents` instead to avoid confusion.

mcgibbon: Sorry, I meant to say "when `subtile_extent` calls `subtile_slice`. subtile_extent returns an int tuple, and "extent" is the correct name for an int tuple describing the domain size.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -141,18 +158,79 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
...
+
+        return tuple(return_extent)
 
     def subtile_slice(
         self,
         rank: int,
         global_dims: Sequence[str],
         global_extent: Sequence[int],
-        overlap: bool = False,
+        overlap: bool = True,
```
mcgibbon, line 233: Please revert this and ask before making new changes to existing APIs.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -166,17 +244,18 @@ def subtile_slice(
...
         return subtile_slice(
-            global_dims,
-            global_extent,
-            self.layout,
-            self.subtile_index(rank),
+            dims=global_dims,
+            global_extent=global_extent,
+            layout=self.layout,
+            subtile_index=self.subtile_index(rank),
+            edge_interior_ratio=self.edge_interior_ratio,
```
mcgibbon, line 258: Good use of keyword arguments!

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -568,23 +647,39 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
             rank_metadata.dims, rank_metadata.extent, self.layout
         )
 
-    def subtile_extent(self, cube_metadata: QuantityMetadata) -> Tuple[int, ...]:
-        """Return the shape of a single rank representation for the given dimensions."""
+    def subtile_extent(
+        self,
+        cube_metadata: QuantityMetadata,
+        rank: int,
+        is_cube_metadata: bool = True,
```
mcgibbon, line 654: Similarly to the other function, please detect the presence of a TILE_DIM directly (keeping the restriction that this must be the first dimension is OK) instead of taking in an `is_cube_metadata` argument.

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -568,23 +647,39 @@ def global_extent(self, rank_metadata: QuantityMetadata) -> Tuple[int, ...]:
...
+
+        return self.tile.subtile_extent(cube_metadata, rank, is_cube_metadata)
 
     def subtile_slice(
         self,
         rank: int,
         global_dims: Sequence[str],
         global_extent: Sequence[int],
-        overlap: bool = False,
+        overlap: bool = True,
```
mcgibbon, line 682: Please revert this and ask before making new changes to existing APIs.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: `is_cube_metadata` has been removed and replaced by a check for a dimension that matches `util.constants.TILE_DIM`.
This check detects the TILE_DIM in any location within the given dimensions, which means that the assertion of it needing to be in the first index can potentially be removed. This assertion is located in `CubedSpherePartitioner.subtile_extent`

```
        if cube_metadata.dims[0] != constants.TILE_DIM:
            raise NotImplementedError(
                "currently only supports tile dimension {constants.TILE_DIM} as the "
                "first dimension, got dims {cube_metadata.dims}"
            )
```

fv3gfs-util/fv3gfs/util/partitioner.py
```diff
@@ -71,7 +71,7 @@ def subtile_slice(
...
```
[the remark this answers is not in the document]
yniederm: Default has been reverted. I'll let you know if there's any further API changes needed.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Done, see the other function's review comment.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: kwargs support + accompanying unit test added.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: I've added some more in-code comments to this block of code, hopefully it's more clear. The two blocks of code do a similar thing, but not exactly. If needed I can probably refactor it slightly into a helper routine.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: I've added some in-line comments to this block of code.

fv3gfs-util/fv3gfs/util/partitioner.py
yniederm: Fair enough. `overlap` isn't needed, you're right.

pace-util/HISTORY.md
```diff
@@ -7,13 +7,16 @@ latest
 Major changes:
 - Renamed DummyComm to LocalComm, and added support for message tags. The DummyComm symbol is still in place for backwards compatibility, but points to LocalComm
 - added error in CubedSphereCommunicator init if given a communicator with a size not equal to the total ranks of the given partitioner
+- `subtile_extent` method of Partitioner classes now takes in a required `rank` argument
+- TilePartitioner has a new `edge_interior_ratio` argument which defaults to 1.0, and lets the user specify the relative 1-dimensional extent of edge and corner compute domains compared to interior compute domains. In all cases, the closest valid value will be used, which enables some previously invalid configurations (e.g. C128 on a 3 by 3 layout will use the closest valid edge_interior_ratio to 1.0)
```
FlorianDeconinck, line 11:
```suggestion
- TilePartitioner has a new `edge_interior_ratio` argument which defaults to 1.0, and lets the user specify the relative 1-dimensional extent of the boundaries (edge and corner) compute domains compared to interior compute domains. In all cases, the closest valid value will be used, which enables some previously invalid configurations (e.g. C128 on a 3 by 3 layout will use the closest valid edge_interior_ratio to 1.0)
```

Trying to align the wording with the code (e,g, `ratio between interior and boundary tile sizes.` in function comment). Boundary is a bit of a loaded term here, so you might want to flip for a constructed "edge-corner" concept or something better.

FlorianDeconinck:
Do we want to have at least one test using a non-default ratio that actually exchange data? Not quite a unit test per se, but it sounds safer to have at least one of those

pace-util/HISTORY.md
mcgibbon: Will do.

mcgibbon:
Good catch, I hadn't noticed there wasn't one. I'll add at least one each of a true MPI and mock MPI case.

pace-util/HISTORY.md
mcgibbon: Done.

FlorianDeconinck (approved):
LGTM!

mcgibbon:
launch jenkins

PR #47, opened 2021-12-04
Add unit tests of DynamicalCore.__call__

mcgibbon:
This PR adds three unit tests of the dynamical core to ensure it does not create new storages or stencils when called, and that it does not access our main sources of global state. The few remaining sources of global state access are removed.

Changes:
- entrypoint.sh now installs pace-util and stencils explicitly, instead of iterating over directories in `external`
- `fv3core.GridIndexing` is now defined
- added `from_f90nml` classmethod constructor to Namelist
- Removed remaining references to `spec.grid` and `global_config` under DynamicalCore
- Added `IntFieldK` to typing
- Added method `set_extra_dim_lengths` to QuantityFactory
- Fixed bug in MetricTerms which would cause it to crash if given a sizer that did not have the required extra_dim_lengths pre-defined


fv3core/fv3core/utils/gt4py_utils.py
```diff
@@ -115,13 +115,6 @@ def make_storage_data(
             else:
                 mask = (n_dims * (True,)) + ((max_dim - n_dims) * (False,))
 
-    if n_dims == 1:
```
mcgibbon, line 118: I expect this breaks the GPU tests. If it does not then that's cool, otherwise I'll have to propagate the backend argument into these helper routines.

mcgibbon: So this does break the tests but not in the way I expected, will revert and propagate the argument.

What this function does in terms of broadcasting dimensions is pretty complex. We might want to avoid using it when we want to do a simple conversion of a numpy array of a given shape into a storage of an identical shape, and use it mainly for converting Fortran data from savepoints into storages of the right shape.

fv3core/docker/entrypoint.sh
```diff
@@ -14,14 +14,8 @@ if [[ -d /external/fv3gfs-wrapper ]]
...
-for package in /external/*
-do
-    if [[ -f $package/setup.py ]]
-    then
-        echo "Setting up $package"
-        pip install -e "$package" -c /constraints.txt
-    fi
-done
+
+pip install -e /external/pace-util -e /external/stencils -c constraints.txt
```
mcgibbon, line 18: I don't know why it triggered in this PR, but without this the packages get installed in the wrong order.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,161 @@
...
+        damping_coefficients=DampingCoefficients.new_from_metric_terms(metric_terms),
+        config=namelist.dynamical_core,
+        phis=state.phis_quantity,
+    )
+    do_adiabatic_init = False
+    # TODO compute from namelist
+    bdt = 225.0
+
+    # TODO: lagrangian contributions currently cause an out of bounds iteration
+    # when halo updates are disabled. Fix that bug and remove this decorator.
```
mcgibbon, line 112: @jdahm , @elynnwu and I tried for a while to fix this bug but concluded this can only happen with an update to gt4py.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def test_call_does_not_access_global_state():
+    dycore, args = setup_dycore()
+
+    def error_func(*args, **kwargs):
+        raise AssertionError("call not allowed")
+
+    mock_grid = unittest.mock.MagicMock()
+    with unittest.mock.patch("fv3core.utils.global_config.get_backend", new=error_func):
```
elynnwu, line 131: Is the next step to eliminate `global_config` all together? Just realized that I could do that in physics today

fv3core/fv3core/_config.py
```diff
@@ -753,6 +753,16 @@ class Namelist:
...
 
+    @classmethod
+    def from_f90nml(cls, namelist: f90nml.Namelist):
+        namelist_dict = namelist_to_flatish_dict(namelist.items())
+        namelist_dict = {
+            key: value
+            for key, value in namelist_dict.items()
+            if key in cls.__dataclass_fields__  # type: ignore
+        }
+        return cls(**namelist_dict)
```
rheacangeo, line 764: side node, we might want to expand Namelist to include some optional parameters, like 'test_case,' the dataclass prevents us from using all of the namelist provided by the namelist_dict, and here expressly removing those extra items

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+
+DIR = os.path.abspath(os.path.dirname(__file__))
+NAMELIST_FILENAME = os.path.join(DIR, "c12_namelist.nml")
+
+
+@contextlib.contextmanager
+def no_lagrangian_contributions(dynamical_core: fv3core.DynamicalCore):
+    # TODO: lagrangian contributions currently cause an out of bounds iteration
+    # when halo updates are disabled. Fix that bug and remove this decorator.
+    # Probably requires an update to gt4py (currently v36).
```
rheacangeo, line 27: it's doing this because the data has non-monotonic pressure. I am not aware of a gt4py change that fixes this, are you? a potential solution (not for this PR) could be for the null communicator to copy columns from the compute domain to the halo of the same rank so that there's real (garbage) data at all points. This workaround looks like it works though.

fv3core/tests/main/test_dycore_call.py
rheacangeo: yes I have the same question -- if we remove the existence of this method, we shouldn't need to test for it, it should be removed. There's the possibility somebody will accidentally bring it back in merging this code with their branch, but we may forget to delete this when we are past the stage of people doing merges. I think that removing global state has been well communicated in the team, so I would think an accidental reverting of code would be caught by the merger, or by the reviewer.

rheacangeo (commented):
This looks great. I made some comments, though the solution may be to address it in a future PR. The main thing I think should be addressed in this PR is code in the repo that is still setting and using the global state. specifically dynamics.py comes to mind, which with your changes here could be updated to not set the global grid, and not use the legacy Grid object.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+
+def setup_dycore() -> Tuple[fv3core.DynamicalCore, List[Any]]:
+    backend = "numpy"
+    stencil_config = fv3core.StencilConfig(
+        backend=backend,
+        rebuild=False,
+        validate_args=True,
+    )
+    mpi_comm = NullComm(rank=0, total_ranks=6, fill_value=0.0)
+    namelist = Namelist.from_f90nml(f90nml.read(NAMELIST_FILENAME))
```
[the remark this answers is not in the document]
mcgibbon: I'll fix this by defining `DynamicalCoreConfig` explicitly in this file.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+    communicator = pace.util.CubedSphereCommunicator(mpi_comm, partitioner)
+    sizer = pace.util.SubtileGridSizer.from_tile_params(
+        nx_tile=namelist.npx - 1,
+        ny_tile=namelist.npy - 1,
+        nz=namelist.npz,
+        n_halo=3,
+        extra_dim_lengths={},
+        layout=namelist.layout,
+        tile_partitioner=partitioner.tile,
+        tile_rank=mpi_comm.rank,
```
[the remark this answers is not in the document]
mcgibbon: Yes, I'll clean something up here. This function should be refactored to just take in `rank` (it does modulo internally already), but not in this PR. This may be something @oelbert  is currently dealing with.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+    grid_indexing = fv3core.GridIndexing.from_sizer_and_communicator(
+        sizer=sizer, cube=communicator
+    )
+    metric_terms = MetricTerms.from_tile_sizing(
```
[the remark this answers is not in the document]
mcgibbon: Yes, ideally I'd like for one QuantityFactory to be created and then passed to everything, which uses it to make both Quantity and storage everywhere. I'll see about the passing-quantity-factory route for this PR, I hadn't noticed a way to do that. I'll definitely be thinking about/trying to simplify this over the next month or so.

fv3core/tests/main/test_dycore_call.py
mcgibbon: I believe `global_config.get_backend` is still used in our runfiles and in our test code. When we remove those functions, we need to be careful to re-run everything.

This test will make sure that even with the code there, it doesn't get called in the course of running the dycore.

So this is the next step, but it's not the current step (for this PR).

mcgibbon:
> The main thing I think should be addressed in this PR is code in the repo that is still setting and using the global state. specifically dynamics.py comes to mind, which with your changes here could be updated to not set the global grid, and not use the legacy Grid object.

Yes, I'd like to make some helper functions that condense the setup logic currently in the test into fewer calls, and then use those functions both in this test and in the dycore runfile. I think this should be done in a later PR.

fv3core/tests/main/c12_namelist.nml
```diff
@@ -0,0 +1,304 @@
+&amip_interp_nml
```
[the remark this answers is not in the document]
rheacangeo: The current namelist differences between the baroclinic and standard cases for low resolution namelists are in "dycore_only," "fdiag", "external_ic", "fv_sg_adj", "na_init", "nggps_ic", "do_ocean", "sfc_override", "sfc_prop_override_nml", "test_case_nml." It is correct these don't currently impact the DynamicalCoreConfig. Specifically fv_sg_adj would turn on fv_subgridz, but is adjacent to the dynamical core but not inside it. We should be careful though with this assumption The performance baroclinic namelists have additional differences. "do_sat_adj" is set to False, turning off a module we run in our standard regression tests, and then "k_split", "n_split", "n_sponge" are set to different values for stability.

fv3core/tests/main/test_dycore_call.py
rheacangeo: there were two things, one we don't have absolute k indexing, which we may have at some point (turbulence needs it), and the other was knowing what the last k index is. But it might be possible to have it passed as an external.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,162 @@
...
+        original_attributes[obj] = obj._lagrangian_contributions
+        obj._lagrangian_contributions = do_nothing  # type: ignore
+    try:
+        yield
+    finally:
+        for obj, original in original_attributes.items():
+            obj._lagrangian_contributions = original
+
+
+def setup_dycore() -> Tuple[fv3core.DynamicalCore, List[Any]]:
```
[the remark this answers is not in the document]
rheacangeo: Ok sure, but at minimum can you test that run_standalone isn't broken by this PR (e.g. by launching one of the performance jenkins plans on your branch)

fv3core/tests/main/c12_namelist.nml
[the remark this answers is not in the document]
mcgibbon: Here I'd like to opt for something that runs as many code pathways as possible, so I'll set do_sat_adj to True and fv_sg_adj to True.

fv3core/tests/main/test_dycore_call.py
mcgibbon: It can currently be passed as an external, and we don't actually need absolute k indexing because of this, but the current version does not support if statements inside for loops inside if statements, or if statements of any kind inside while loops, and various other combinations that could make this possible.

fv3core/tests/main/test_dycore_call.py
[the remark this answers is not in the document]
mcgibbon: Updated. There was a typo in the MetricTerms init (setting extra_dim_lenths instead of extra_dim_lengths) so I made this safer using a new public method (instead of also setting a private attribute).

mcgibbon:
Performance tests passed without runfile update ([log](https://jenkins.ginko.ch/view/pace/job/pace-fv3core-performance-test/55/)), waiting on results with set_grid removed ([log](https://jenkins.ginko.ch/view/pace/job/pace-fv3core-performance-test/56/)).

mcgibbon:
launch jenkins

fv3core/tests/main/test_dycore_call.py
rheacangeo: I think Johann added that capability already. You need to know the current absolute k index to be able to compare it to the external max allowed k. Maybe that is also possible now.

fv3core/tests/main/test_dycore_call.py
```diff
@@ -0,0 +1,201 @@
...
+    backend = "numpy"
+    stencil_config = fv3core.StencilConfig(
+        backend=backend,
+        rebuild=False,
+        validate_args=True,
+    )
+    mpi_comm = NullComm(rank=0, total_ranks=6, fill_value=0.0)
```
rheacangeo, line 57: if you define this after the config, you could use the layout to determine the total_ranks. not a required change since it is test code

rheacangeo (approved):
One minor comment, but can be merged as-is if preferred (once it passes)

fv3core/tests/main/test_dycore_call.py
mcgibbon: Johann was helping us fix it :) Apparently he has implemented these capabilities, they just aren't in v36.

fv3core/tests/main/test_dycore_call.py
mcgibbon: Good idea, in the end I'll avoid doing this only because in practice we'll want to either configure layout from the MPI comm or validate layout from the MPI comm, and this example might confuse someone into writing it the other way around.

mcgibbon: Wait my comment makes no sense, in this production example it's not possible to make an MPI comm! So I'll put this in.

rheacangeo:
launch jenkins

mcgibbon:
launch jenkins

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

PR #49, opened 2021-12-06
Fix/enable physics linting

elynnwu:
Some more cleanup:
- Delete global_config in physics, use shared float field instead
- Typo and documentation cleanup


fv3gfs-physics/setup.py
```diff
@@ -30,8 +30,8 @@
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
     ],
-    description="fv3gfs-physics is a gt4py-based physical parameterization \
-        for atmospheric models",
+    description="fv3gfs-physics is a gt4py-based physical parameterization "
+    "for atmospheric models",
```
mcgibbon, line 34: Does this work without the parentheses surrounding the two strings?

mcgibbon (approved):
Thanks!

PR #51, opened 2021-12-08
physics and dycore share constants

rheacangeo:
- Move fv3core.util.global_constants and fv3gfs-physics.global_constants to pace.util.constants
 - physics and dycore overlapping constants consolidated -- mainly duplication of constants in saturation adjustment and microphysics removed, and both use pace.util.constants instead
 - physics constants made upper case
 - physics constants that come from the namelist are added as externals to microphysics stencil, and added to the NamelistDefaults and Namelist class (still in fv3core, but will presumably move up). Physics may need a PhysicsConfig to build from this, but using a full Namelist object should work. 
 - moved a couple of Microphysics call time constants to the stencil externals as well
 - Microphysics-only constants (verified in fortran they only show up in the gfdl microphysics scheme) are moved in the microphysics_functions.py -- they are easy to move back into pace.util.constants if that is preferred


fv3core/fv3core/_config.py
```diff
@@ -64,6 +64,50 @@ class NamelistDefaults:
     nf_omega = 1
     fv_sg_adj = -1
     n_sponge = 1
+    # all below added for physics
```
mcgibbon, line 67: nit: it's going to be very easy for someone to add a non-physics value below, so you might want to consider removing the comment.

```suggestion
    # all below are for physics
```

mcgibbon (approved):
Very nice!

fv3core/fv3core/_config.py
rheacangeo: k thanks, I'll remove the comment

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

fv3core/fv3core/_config.py
```diff
@@ -553,41 +597,47 @@ class Namelist:
...
+    vg_fac: float = NamelistDefaults.vg_fac
+    vi_fac: float = NamelistDefaults.vi_fac
+    vr_fac: float = NamelistDefaults.vr_fac
+    de_ice: bool = NamelistDefaults.de_ice
+    do_qa: bool = NamelistDefaults.do_qa
+    do_sedi_heat: bool = NamelistDefaults.do_sedi_heat
+    do_sedi_w: bool = NamelistDefaults.do_sedi_w
+    fast_sat_adj: bool = NamelistDefaults.fast_sat_adj
+    fix_negative: bool = NamelistDefaults.fix_negative
+    irain_f: float = NamelistDefaults.irain_f
```
elynnwu, line 618: This one should be a boolean too

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -653,7 +667,7 @@ def warm_rain(
 
         # Auto-conversion assuming linear subgrid vertical distribution of
         # cloud water following lin et al. 1994, mwr
-        if constants.irain_f != 0:
+        if irain_f != 0:
```
elynnwu, line 670: Change to boolean

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -794,7 +808,7 @@ def sedimentation(
 
         with interval(0, 1):
 
-            if tz > constants.tice:
+            if tz > tice:
                 stop_k = 1
```
elynnwu, line 812: So we cannot have temporary boolean right? Is this something that we should support when we add typed temporaries?

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -924,60 +938,44 @@ def sedimentation(
...
         with interval(1, -1):
 
-            if (
-                (constants.vi_fac >= 1.0e-5)
-                and (no_fall[0, 0, -1] == 0)
-                and (zt >= zt[0, 0, -1])
-            ):
-                zt = zt[0, 0, -1] - constants.dz_min
+            if (vi_fac >= 1.0e-5) and (no_fall[0, 0, -1] == 0) and (zt >= zt[0, 0, -1]):
+                zt = zt[0, 0, -1] - functions.DZ_MIN_FLIP
```
elynnwu, line 942: I think this is okay because `DZ_MIN_FLIP` is just a constant used inside microphysics functions. But I was confused in the first place that this is a function. This is optional - but maybe we should change the import package as name to something other than `functions`?

elynnwu: It is also true that constants are all caps, so this should also fine as it is

elynnwu (approved):
Thanks for going through this! I think there's just one namelist variable that should be a boolean, and that should be it for this PR.

fv3gfs-physics/fv3gfs/physics/functions/microphysics_funcs.py
```diff
@@ -260,123 +294,124 @@ def revap_racc(
...
-                * (tc * (constants.aa * tc + constants.bb) + constants.cc)
-                + constants.dd_fs * tc
-                + constants.ee
+                (3.0 + log(qi * den) / log_10) * (tc * (AA * tc + BB) + CC)
+                + DD_FS * tc
+                + EE
             )
             vti = vi0 * exp(log_10 * vti) * 0.8
-            vti = min(constants.vi_max, max(constants.vf_min, vti))
+            vti = min(vi_max, max(VF_MIN, vti))
```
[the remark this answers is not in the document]
rheacangeo: maybe it should be in the namelist? it wasn't in the fortran, but seems like it would make sense to

fv3gfs-physics/fv3gfs/physics/functions/microphysics_funcs.py
```diff
@@ -260,123 +294,124 @@ def revap_racc(
...
-        if qi < constants.thi:
+        if qi < THI:
 
-            vti = constants.vf_min
+            vti = VF_MIN
 
         else:
 
-            tc = tk - constants.tice
+            tc = tk - tice
```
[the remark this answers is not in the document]
rheacangeo: it's very weird. in fortran it is tice vs t_ice

fv3core/fv3core/_config.py
rheacangeo: that one is set to a integer in fortran (though the name sounds like it should be a float haha)

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
rheacangeo: I think people have namelists with this as 0 or 1

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
rheacangeo: we can, it'll just be a 3d boolean

PR #56, opened 2021-12-09
Gt4py is used as a submodule

mcgibbon:
This PR updates pace to use gt4py pinned as a submodule, rather than using GT4PY_VERSION.txt. After this PR, the gt4py caches will use the gt4py sha rather than tag in their GT4PY_VERSION.txt files.


mcgibbon:
launch jenkins

mcgibbon:
launch jenkins

docker/Makefile
```diff
@@ -3,35 +3,24 @@ SHELL := /bin/bash
...
-.PHONY: dev fv3gfs_image build build_deps push_deps pull_deps get_gt4py
-
-get_gt4py:
-	if [ ! -f ${GT4PY_DIR}/pyproject.toml ]  || \
-	[ "$(shell cd ${GT4PY_DIR} && git describe --tags )" != "${GT4PY_VERSION}"  ]; then \
-                rm -rf ${GT4PY_DIR} && \
-	  	git clone https://github.com/ai2cm/gt4py.git ${GT4PY_DIR} && \
-	  	cd ${GT4PY_DIR} && git checkout ${GT4PY_VERSION};\
-	fi
+.PHONY: dev fv3gfs_image build build_deps push_deps pull_deps
```
eddie-c-davis, line 20: So much cleaner!

.gitmodules
```diff
@@ -4,6 +4,9 @@
 [submodule "external/fv3gfs-wrapper"]
 	path = external/fv3gfs-wrapper
 	url = git@github.com:ai2cm/fv3gfs-wrapper.git
+[submodule "external/gt4py"]
```
eddie-c-davis, line 7: Is this where specific tags would be indicated (e.g., the sha for `v36`)?

external/daint_venv/install.sh
```diff
@@ -37,17 +37,6 @@ if [ $save_wheel ]; then
 fi
 python3 -m pip install --find-links=$wheeldir cupy Cython clang-format
 
-
-# installation of gt4py
```
eddie-c-davis, line 41: Nice to see this going away.

mcgibbon:
Launched performance test as https://jenkins.ginko.ch/job/pace-fv3core-performance-test/65/

fv3core/external/gt4py
```diff
@@ -0,0 +1 @@
+../../external/gt4py
```
eddie-c-davis, line 1: Is this one of the symlinks we have discussed?

pace-util/constraints.txt
```diff
@@ -1,100 +1,338 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.8
```
eddie-c-davis, line 2: Much easier on the eyes...

eddie-c-davis (approved):
This is a heroic effort  that simplifies the repository considerably. Have a few questions, but LGTM!

mcgibbon:
launch jenkins

rheacangeo:
What is the plan if we want to run an older commit on pace, but the gt4py-develop commit-hash no longer exists? The previous process involved an automatic tag that would preserve it, will that be a manual process now?

mcgibbon:
> What is the plan if we want to run an older commit on pace, but the gt4py-develop commit-hash no longer exists? The previous process involved an automatic tag that would preserve it, will that be a manual process now?

The plan is to still make commit tags to preserve commits.

rheacangeo:
But it will no longer be automatic, this is a little risky

fv3core/external/gt4py
mcgibbon: I'm not positive I know what you mean, but i think yes?

.gitmodules
mcgibbon: You'd indicate a tag by checking out that commit in the submodule, and adding it to the repo with `git add external/gt4py` or similar. Checking it out by tag or commit makes no difference.

mcgibbon:
Let's continue the discussion at https://github.com/ai2cm/pace/pull/64.

PR #57, opened 2021-12-09
update_dwind_phys 54 ranks fix

rheacangeo:
fixing npx and npy to be tile values instead of local ones so that the update_dwind_phys tests works for 54 ranks


elynnwu (approved):
lgtm!

PR #58, opened 2021-12-09
stop re-downloading of data on gsutil fail

rheacangeo:
use gsutil cp - rather than gsutil cat because it has a checksum and safety, while gsutil cat is apparently the wild west.


elynnwu (approved):

rheacangeo:
launch jenkins

mcgibbon:
launch jenkins

rheacangeo:
launch jenkins

PR #60, opened 2021-12-10
Remove global access in fv3core and physics tests

elynnwu:
This PR is a prerequisite to moving namelist to driver. The goal for this PR is to 1) remove global config access, 2) share translate class infrastructure, 3) remove dependencies to `fv3core` in physics tests.

Changes:
- Move `grid`, `global_config`, and `corners` from `fv3core.utils` to `pace.util.testing` and `pace.util`: global_config is removed from most places and can be deleted soon-ish
- Move `translate` and `parallel_translate` to `pace.util.testing`: they are shared in dycore and physics
- Move `NamelistDefaults` to `pace.util.namelist`
- Add `PhysicsConfig` class: when namelist is moved to the driver, we should add a method to return physics config
- Add filter in physics savepoint test to only check for physics tests: since the logic checks for every serialized savepoint, pytest also tries to run dycore tests (with no matching test in the physics folder). Adding this avoid checking all serialized savepoints, but future physics test needs to be included here.
- Pass `namelist` and `stencil_factory` down to each savepoint class
- Remove `stencil_facotry` from Grid class


fv3gfs-physics/fv3gfs/physics/testing/translate.py
```diff
@@ -0,0 +1,354 @@
...
+                self.data[key] = utils.make_storage_data(
+                    value,
+                    shape,
+                    origin=origin,
+                    start=origin,
+                    read_only=True,
+                    backend=self.backend,
+                )
+
+    def python_grid(self):
```
elynnwu, line 350: All `fv3core` import in physics test have been removed except for `Grid` here. Should we make a new classmethod in `GridData` that makes the class with serialized data? Or should we retain `Grid`? We also need a location for either of them.

fv3gfs-physics/fv3gfs/physics/_config.py
```diff
@@ -0,0 +1,125 @@
...
+        if isinstance(value, dict):
+            for subkey, subvalue in value.items():
+                if subkey in flatter_namelist:
+                    raise ValueError(
+                        "Cannot flatten this namelist, duplicate keys: " + subkey
+                    )
+                flatter_namelist[subkey] = subvalue
+        else:
+            flatter_namelist[key] = value
+    return flatter_namelist
```
mcgibbon, line 125: `namelist_to_flatish_dict` should probably go to `util` since it's used identically in the dycore. This can also live in `pace.util.namelist`.

fv3gfs-physics/fv3gfs/physics/testing/translate.py
mcgibbon: Grid's main responsibility is to convert serialized data, which may or may not include some number of halos and may or may not be staggered, into standardized storages that all have 3 halos and a compute origin at (3, 3). For this reason, Grid should continue to be used in the test code, and should probably live with the translate infrastructure.

fv3gfs-physics/tests/savepoint/translate/translate_fv_update_phys.py
```diff
@@ -34,9 +33,11 @@ def __getitem__(self, item):
 
 
 class TranslateFVUpdatePhys(ParallelTranslate2Py):
-    def __init__(self, grids):
-        super().__init__(grids)
+    def __init__(self, grids, namelist, stencil_factory):
+        super().__init__(grids, namelist, stencil_factory)
```
mcgibbon, line 37: Technically it shouldn't be necessary to pass `stencil_factory` separate from `grids`, since each `grid` object has its own stencil factory. It would be different if we are able to remove `grid`, it also would be different if we could remove the `Grid.stencil_factory` attribute and _only_ make stencil factory directly.

Do you think for now we should remove the stencil_factory argument? I can see keeping it if you think this is a step on a road towards retrieving fewer attributes from Grid and eventually being able to use it only for array transformations. I'm kind of undecided on this one.

fv3gfs-physics/tests/savepoint/conftest.py
```diff
@@ -48,8 +57,7 @@ def data_path(pytestconfig):
 def data_path_from_config(config):
     data_path = config.getoption("data_path")
     namelist_filename = os.path.join(data_path, "input.nml")
-    fv3core._config.set_namelist(namelist_filename)
-    return data_path
+    return data_path, namelist_filename
```
mcgibbon, line 60: Please update the function name, and possibly also add a return value type hint.

fv3gfs-physics/tests/savepoint/conftest.py
```diff
@@ -99,12 +111,32 @@ def read_serialized_data(serializer, savepoint, variable):
     return data
 
 
-def process_grid_savepoint(serializer, grid_savepoint, rank, *, backend: str):
-    grid = make_grid(grid_savepoint, serializer, rank, backend=backend)
-    fv3core._config.set_grid(grid)
+def Grid(serializer, grid_savepoint, rank, layout, *, backend: str):
```
mcgibbon, line 114: Please use a lowercase name, since this isn't returning a class.

fv3gfs-physics/tests/savepoint/conftest.py
```diff
@@ -213,16 +264,18 @@ def sequential_savepoint_cases(metafunc, data_path, *, backend: str):
...
 
     # Set the grid to rank 0's data
     serializer = get_serializer(data_path, 0)
     grid_savepoint = serializer.get_savepoint(GRID_SAVEPOINT_NAME)[0]
-    grid_rank0 = process_grid_savepoint(serializer, grid_savepoint, 0, backend=backend)
-    fv3core._config.set_grid(grid_rank0)
-
+    grid_rank0 = Grid(
+        serializer, grid_savepoint, 0, physics_config.layout, backend=backend
+    )
```
mcgibbon, line 278: This variable is unused, you should be able to delete it.

mcgibbon:
I’m a little worried that if we duplicate the base translate classes and make small changes, these classes are very complex and may be hard to merge later. Would it be possible to put these in `pace.util.testing` in a way that both physics and dycore continue to use the same code?

fv3gfs-physics/fv3gfs/physics/_config.py
elynnwu: Moved to `pace.util.namelist`

fv3gfs-physics/tests/savepoint/conftest.py
elynnwu: Changed to `upper_grid`

fv3gfs-physics/fv3gfs/physics/testing/translate.py
elynnwu: This makes sense, I had thought that we wanted to get rid of `Grid` all together. If the plan is to keep it alive within the translate infrastructure, I'll go ahead and reshuffle some things.

fv3gfs-physics/tests/savepoint/translate/translate_fv_update_phys.py
elynnwu: Yes, let's remove the `stencil_factory` argument and access it through grid.

fv3core/tests/main/test_config.py
```diff
@@ -17,12 +17,12 @@
...
-    new_backend = "new_backend"
-    assert new_backend != start_backend
-    fv3core.set_backend(new_backend)
-    assert fv3core.get_backend() == new_backend
+# def test_set_backend():
+#     start_backend = fv3core.get_backend()
+#     new_backend = "new_backend"
+#     assert new_backend != start_backend
+#     fv3core.set_backend(new_backend)
+#     assert fv3core.get_backend() == new_backend
```
rheacangeo, line 25: this can be deleted now?

fv3core/tests/main/test_grid.py
```diff
@@ -262,7 +262,7 @@ def test_domain_compute(
     add: Index3D,
     domain_compute: Index3D,
 ):
-    grid = fv3core.utils.grid.GridIndexing(
+    grid = pace.stencils.testing.grid.GridIndexing(
```
rheacangeo, line 265: is GridIndexing in pace.dsl or pace.stencils.testing.grid?

rheacangeo (commented):
Epic. As discussed, there are some known circular dependencies that come about from this. But we agree, this PR is large enough, and if everything works, we should merge this and make a new story to address those smaller cycles.

fv3core/fv3core/stencils/delnflux.py
```diff
@@ -13,9 +13,9 @@
 )
 
 import pace.dsl.gt4py_utils as utils
-from fv3core.utils.grid import DampingCoefficients, axis_offsets
 from pace.dsl.stencil import StencilFactory, get_stencils_with_varied_bounds
 from pace.dsl.typing import FloatField, FloatFieldIJ, FloatFieldK
+from pace.stencils.testing.grid import DampingCoefficients, axis_offsets
```
mcgibbon, line 18: I see most stencil files import from pace.stencils.testing.grid, is there a way to move whatever they’re using out of testing while keeping our testing-only files there? GridData, axis_offsets, and DampingCoefficients don't really belong in `testing`.

stencils/pace/stencils/testing/grid.py
```diff
@@ -1,16 +1,16 @@
...
 
 import numpy as np
 from gt4py import gtscript
 
-import fv3core.utils.global_config as global_config
 import pace.util
-from fv3core.grid import MetricTerms
 from pace.dsl import gt4py_utils as utils
-from pace.dsl.stencil import GridIndexing, StencilConfig, StencilFactory
+from pace.dsl.stencil import GridIndexing
```
mcgibbon, line 11: Ideally only Grid should remain in this file, the other things should be moved out of `testing`. If we no longer pass `Grid` to `axis_offsets`, you should be able to modify it to only accept GridIndexing, which will allow you to move it out of this file. If we do still pass `Grid` to `axis_offsets`, you can leave the legacy function in this file and call it from whichever testing code is using it.

fv3core/fv3core/testing/translate_fvdynamics.py
```diff
@@ -270,6 +269,8 @@ def __init__(self, grids, *args, **kwargs):
             self.ignore_near_zero_errors[qvar] = True
         self.ignore_near_zero_errors["q_con"] = True
         self.dycore: Optional[fv_dynamics.DynamicalCore] = None
+        self.stencil_factory = stencil_factory
+        self.namelist = namelist
```
mcgibbon, line 273: Can you type hint `namelist`, since this is a very overloaded term?

fv3core/tests/main/test_config.py
mcgibbon, line 25:
```suggestion
```

fv3core/tests/main/test_dycore_call.py
```diff
@@ -170,9 +170,9 @@ def error_func(*args, **kwargs):
         raise AssertionError("call not allowed")
 
     mock_grid = unittest.mock.MagicMock()
-    with unittest.mock.patch("fv3core.utils.global_config.get_backend", new=error_func):
+    with unittest.mock.patch("pace.util.global_config.get_backend", new=error_func):
         with unittest.mock.patch(
-            "fv3core.utils.global_config.is_gpu_backend", new=error_func
+            "pace.util.global_config.is_gpu_backend", new=error_func
```
mcgibbon, line 175: Make sure there's a Jira task to delete pace.util.global_config, if we're no longer using it.

fv3core/tests/savepoint/conftest.py
```diff
@@ -99,12 +100,24 @@ def read_serialized_data(serializer, savepoint, variable):
     return data
 
 
-def process_grid_savepoint(serializer, grid_savepoint, rank, *, backend: str):
-    grid = make_grid(grid_savepoint, serializer, rank, backend=backend)
-    fv3core._config.set_grid(grid)
+def upper_grid(serializer, grid_savepoint, rank, layout, *, backend: str):
```
mcgibbon, line 103: What is an "upper" grid? Is there a reason for this function when you can directly call `make_grid`?

fv3gfs-physics/fv3gfs/physics/testing/__init__.py
```diff
@@ -1,2 +0,0 @@
-from . import translate_physics
-from .translate_physics import TranslatePhysicsFortranData2Py
```
mcgibbon, line 2: Why were these removed?

fv3gfs-physics/tests/savepoint/conftest.py
```diff
@@ -23,9 +24,18 @@
...
+    "GFSPhysicsDriver",
+    "AtmosPhysDriverStatein",
+    "PrsFV3",
+    "PhiFV3",
+    "Microph",
+    "FillGFS",
+    "PhysUpdateTracers",
+    "PhysUpdatePressureSurfaceWinds",
+    "UpdateDWindsPhys",
+    "FVUpdatePhys",
```
mcgibbon, line 37: Cool to have a list of these!

fv3gfs-physics/tests/savepoint/conftest.py
```diff
@@ -99,12 +112,32 @@ def read_serialized_data(serializer, savepoint, variable):
     return data
 
 
-def process_grid_savepoint(serializer, grid_savepoint, rank, *, backend: str):
-    grid = make_grid(grid_savepoint, serializer, rank, backend=backend)
-    fv3core._config.set_grid(grid)
+def upper_grid(serializer, grid_savepoint, rank, layout, *, backend: str):
+    grid = make_grid(grid_savepoint, serializer, rank, layout, backend=backend)
```
mcgibbon, line 116: Can you call `make_grid` directly?

I'm a little confused by this network of functions (`grid_data`, `grid_indexing`, `physics_config`, etc.). Instead of calling `grid_data(grid)`, should you directly use `grid.grid_data` in whatever function is calling these?

mcgibbon (commented):
Very nice seeing the test code make direct use of `namelist` and `stencil_factory` instead of relying on `Grid` for everything.

You can leave refactoring `pace.stencils.testing.grid` for a follow-on PR, but please make a Jira task. It's not great having every stencil file import from `testing`, since the idea of putting things there is to indicate the stencil code shouldn't be importing from it.

Very few comments for how much you've updated. Great job!

fv3core/examples/standalone/runfile/dynamics.py
```diff
@@ -227,32 +231,36 @@ def collect_data_and_write_to_file(
             profiler = cProfile.Profile()
             profiler.disable()
 
-        fv3core.set_backend(args.backend)
-        fv3core.set_rebuild(False)
-        fv3core.set_validate_args(False)
```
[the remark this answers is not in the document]
elynnwu: great catch, changed to False

fv3core/examples/standalone/runfile/dynamics.py
```diff
@@ -227,32 +231,36 @@ def collect_data_and_write_to_file(
...
 
         # TODO remove this creation of the legacy grid once everything that
         # references it is updated or removed
-        grid = spec.make_grid_from_namelist(namelist, rank)
-
+        grid = spec.make_grid_from_namelist(dycore_config, rank, args.backend)
+        stencil_config = pace.dsl.stencil.StencilConfig(
+            backend=args.backend,
+            rebuild=False,
+            validate_args=True,
```
rheacangeo, line 254: change to False please

fv3core/examples/standalone/runfile/dynamics.py
```diff
@@ -227,32 +231,36 @@ def collect_data_and_write_to_file(
...
+        partitioner = util.CubedSpherePartitioner(
+            util.TilePartitioner(dycore_config.layout)
+        )
         communicator = util.CubedSphereCommunicator(mpi_comm, partitioner)
 
         # TODO remove this creation of the legacy grid once everything that
         # references it is updated or removed
-        grid = spec.make_grid_from_namelist(namelist, rank)
-
+        grid = spec.make_grid_from_namelist(dycore_config, rank, args.backend)
```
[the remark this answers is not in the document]
elynnwu: `stencil_factory` is needed for the baroclinic test case too, so can't be inside `read_serialized_initial_state` only

fv3core/tests/main/test_grid.py
elynnwu: It is `pace.dsl`, updated here and elsewhere

.pre-commit-config.yaml
```diff
@@ -33,13 +31,42 @@ repos:
           --follow-imports, skip  # needed so we only test enabled files
         ]
       files: pace-util
+    - id: mypy
```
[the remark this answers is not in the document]
elynnwu: Good idea! This fixed one of my problems too, I couldn't figure out it kept running on a file I excluded in stencils, turns out I had 2 mypy stencils and the later one didn't have the exclude

examples/dyn_phy.py
```diff
@@ -2,6 +2,7 @@
 import time
 
 import click
+import f90nml
```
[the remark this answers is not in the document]
elynnwu: Yes, we should add a plan for this. This script was actually broken for a while and we didn't notice.

fv3core/fv3core/_config.py
```diff
@@ -969,7 +857,7 @@ def set_grid(in_grid):
     grid = in_grid
 
 
-def set_namelist(filename, rank=0):
+def set_namelist(filename, backend, rank=0):
```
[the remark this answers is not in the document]
elynnwu: Great catch, this one is not called anywhere anymore! Deleting `set_namelist`

fv3core/tests/main/test_config.py
elynnwu: Rhea noticed the same thing too 👍

fv3core/tests/main/test_dycore_call.py
elynnwu: It's being used in two spots in the main tests. Once we get rid of those, we can delete `global_config`

fv3core/tests/savepoint/conftest.py
elynnwu: It was originally going to be a pytest fixture, but you're right we can just call make_grid directly. Changing this.

fv3gfs-physics/tests/savepoint/conftest.py
elynnwu: They were supposed to be pytest fixtures, but it's easier to make them instead. I forgot to clean up, deleting all these now.

fv3gfs-physics/fv3gfs/physics/testing/__init__.py
elynnwu: Since we no longer need to carry separate translate class infrastructure, we don't need this in physics anymore. If we decide to add `TranslatePhysics`, it will go here.

rheacangeo (approved):
Wahoo! I think the Grid object served us well, but good to finalize the shift away from it by untangling the objects we were grabbing from it as an interim step. Lots going on here and this will make many things possible!

mcgibbon (approved):
Make sure there are the 2 Jira tasks for removing global_config and for moving utilities out of `pace.stencils.testing.grid`, otherwise looks good to me!

PR #61, opened 2021-12-10
changed_from_main can compare against a git hash

rheacangeo:
changed_from_main can optionally compare to a git hash rather than the main branch. main is default. This will allow us to only run the tests on code that changed when main is updated, and compare to GIT_PREVIOUS_SUCCESSFUL_COMMIT

### DIFF
--- a/changed_from_main.py
+++ b/changed_from_main.py
@@ -70,6 +70,12 @@ def parse_args(subdir_dependencies: Dict[str, Any]):
         help="subdirectory name of project to check",
         choices=subdir_dependencies.keys(),
     )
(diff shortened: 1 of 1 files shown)


elynnwu (approved):
Thanks for adding this Rhea! Definitely better than the hacky solutions we talked about 😄

rheacangeo:
launch jenkins

rheacangeo:
While I think this is a reasonable feature for this script to have, for the script to not be so hard-coded to 'main,' we decided we don't trust the script to catch all the cases it needs to on merges to main, so let's just not do this.

PR #65, opened 2021-12-14
parallel and sequential tests cache compiled stencil to the same folder

rheacangeo:
setting the gtscript.stencil "name" argument to the funcs module + name, so that whether or not a future stencil is used, the compiled stencil ends up in the same cache file, so we don't have to run multiple versions of the code to generate the cache, and we can keep the file count low (or the daint inode limit).


eddie-c-davis (approved):
Looks good to me, great catch!

PR #66, opened 2021-12-15
Run pace-util tests on gpu on CircleCI

mcgibbon:


mcgibbon:
Turns out we don't have access to gpu executors. Closing PR.

PR #67, opened 2021-12-15
init baroclinic gpu

rheacangeo:
On GPU the initilized state was actually all 0s, as the effort to keep the code from crashing on GPU actually entirely separated the gpu arrays from the cpu arrays. It is possible to pass cupy arrays (the state.quantity.data instead of the state.storage.data, which is a memory view and cannot be sliced) in and still do it this way, but this requires modifying almost all the lines of math to be cupy friendly (e.g. cp.asarray wrappers around many statements). Instead, we compute on the cpu in numpy, and from that create a DycoreState. 
This means we won't use the quantity.view feature to slice unless we revisit using a a DycoreState initialized with 0's to start with

### DIFF
--- a/fv3core/fv3core/initialization/baroclinic.py
+++ b/fv3core/fv3core/initialization/baroclinic.py
@@ -425,8 +425,8 @@ def init_baroclinic_state(
     Create a DycoreState object with quantities initialized to the Jablonowski &
     Williamson baroclinic test case perturbation applied to the cubed sphere grid.
     """
-
-    shape = (*metric_terms.lat.data.shape[0:2], metric_terms.ak.data.shape[0])
+    sample_quantity = metric_terms.lat
+    shape = (*sample_quantity.data.shape[0:2], metric_terms.ak.data.shape[0])
     nx, ny, nz = local_compute_size(shape)
     numpy_dict = {}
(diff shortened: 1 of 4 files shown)


rheacangeo:
launch jenkins

fv3core/fv3core/initialization/baroclinic.py
```diff
@@ -423,63 +425,65 @@ def init_baroclinic_state(
     Create a DycoreState object with quantities initialized to the Jablonowski &
     Williamson baroclinic test case perturbation applied to the cubed sphere grid.
     """
-
-    state = DycoreState.init_empty(metric_terms.quantity_factory)
-    nx, ny, nz = local_compute_size(state.delp.data.shape)
+    sample_quantity = metric_terms.lat
```
twicki, line 428: can we extract the initialization of the numpy state out in it's own function?

twicki (changes requested):

fv3core/fv3core/initialization/baroclinic.py
rheacangeo: yes! good idea

twicki (approved):
looks great

PR #70, opened 2021-12-17
Allow relative path in install virtualenv

elynnwu:
We currently only support absolute path for installing virtualenv. In this PR, we allow both absolute and relative path as an input to install_virtualenv.


twicki (approved):
lgtm

PR #71, opened 2021-12-17
ensure that artifacts are stored if in the main branch

twicki:
We only saved artifacts from `origin/master` - this led to no performance data saved after switching to pace


rheacangeo (approved):
Thank you!

PR #79, opened 2021-12-20
add pace driver, update dycore and physics state

rheacangeo:
## Purpose

Added a 'pace-driver' package for use in managing model workflows. 

## Code changes
* Added a simple Driver class that is meant as a stepping-stone, not the final design, to move the complexity of objects being created in runfiles into the 'model.' 
* The driver can use serialized initial state data or baroclinic
* Updated the PhysicsState to use similar conventions as the DycoreState, and include metadata that indicates whether it is shared with dycore (full_model_var). In this version shared variables are not copied, but shared. We may want to adjust and pull up stencil code that transforms dycore variables to physics variables and vice versa up to the Driver level, but that is likely better as a future iteration.
* MicrophysicsState doesn't initialize storages in call-time
* Moved fv_update_phys, update_atmos_state, and update_dwind_phys into pace-stencils, and the state updater object into the Driver. 
* u_dt, v_dt, and t_dt in the updater are quantities
* remove 'partitioner' as a separate argument from the communicator in Physics classes because we can get the partitioner from the communicator.
* created a DriverGridData class that extracts the variables update_dwind_phys needs from MericTerms. We no longer need serialized grid data for pace (except for testing regressions purposes). 
* Physics no longer refers to DycoreState
* removed hard-coded ptop from Physics -- it is available in grid_data
* added examples/run_driver.py as a script that demonstrates the Driver class. We could alternatively adapt examples/dyn_phy.py and remove this file. 
* Namelist gets the dycore_only configuration option turned on so we can use it. 
* MetricTerms has a new method for splitting cartesian variables into FloatIJ storages, as seems like it could be a common use-case for someone using MetricTerms.
* Updated test code to go with the changes

## Infrastructure changes
* 'driver' at the top level provides the pace.dsl package
*  added pace.driver to Dockerfile, entrypoint
* added driver tests


fv3gfs-physics/fv3gfs/physics/physics_state.py
```diff
@@ -1,120 +1,261 @@
...
+        metadata={"name": "interface_pressure", "units": "Pa", "full_model_var": False}
+    )
+    prsik: FloatField = field(
+        metadata={
+            "name": "log interface_pressure",
+            "units": "Pa",
+            "full_model_var": False,
+        }
+    )
+    quantity_factory: InitVar[pace.util.QuantityFactory]
```
mcgibbon, line 251: Can you remove the `quantity_factory` attribute? It doesn't get used in this PR, and it can change the meaning of the PhysicsState fairly significantly (from a container of data, to a container of data and an operator).

mcgibbon (changes requested):
Very cool! My main comments are:
- We'll need to think about some of the design for the Driver, but this is a good starting point
- Some tests will make sure we don't break this code, and that its features work as intended
- We should have another group chat about type hints on public functions/methods - with pylance, whether these are present or not makes a big difference for me as a coderd

mcgibbon:
You may be able to simplify testing into a unit test by mocking out the DynamicalCore and corresponding Physics class, replacing them with mocks that check they were called.

stencils/pace/stencils/testing/grid.py
```diff
@@ -1089,3 +1089,93 @@ def _grid_indexing_axis_offsets(
     domain: Tuple[int, ...],
 ) -> Mapping[str, gtscript.AxisIndex]:
     return grid.axis_offsets(origin=origin, domain=domain)
+
+
+@dataclasses.dataclass(frozen=True)
+class DriverGridData:
```
elynnwu, line 1095: This is great, now we can retain the ability in translate test to read these from serialized data. Also, this should make it easier to add these variables in metric terms.

elynnwu (commented):
This is a great start!

driver/pace/driver/driver.py
```diff
@@ -0,0 +1,167 @@
...
+        metric_terms = MetricTerms(
+            quantity_factory=self.quantity_factory, communicator=self.comm
+        )
+        grid_data = GridData.new_from_metric_terms(metric_terms)
+        return metric_terms, grid_data
+
+    def initialize_dycore_state(self, dycore_init_mode):
+        if dycore_init_mode == "serialized_data":
+            sys.path.append("/usr/local/serialbox/python/")
+            import serialbox
```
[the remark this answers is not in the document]
mcgibbon: We'd like serialbox to be pip installable, but there's a bug in serialbox that still prevents this.

driver/pace/driver/driver.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(
+        self,
+        namelist,
+        comm,
+        backend,
+        dycore_init_mode="serialized_data",
+        rebuild=False,
+        validate_args=True,
+        data_dir=None,
+    ):
```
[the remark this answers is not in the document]
rheacangeo: comm is specifically something we don't have a good typehint for yet. I added a docstring

driver/pace/driver/driver.py
```diff
@@ -0,0 +1,167 @@
...
+from fv3core.testing.translate_fvdynamics import init_dycore_state_from_serialized_data
+from fv3gfs.physics import PhysicsState
+from fv3gfs.physics.stencils.physics import Physics
+from pace.stencils.testing.grid import DampingCoefficients, DriverGridData, GridData
+from pace.stencils.update_atmos_state import UpdateAtmosphereState
+from pace.util.constants import N_HALO_DEFAULT
+from pace.util.grid import MetricTerms
+
+
+class Driver:
```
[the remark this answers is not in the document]
rheacangeo: I added some very minimal tests. I spent quite a bit of time to get a run where I could compare data changes due to dynamics and physics, but it is a bit blocked on a number of fronts. It worked inside the container, but  testing outside the container sometimes resulted in compilation errors. When I use the NullComm, all the data are nan's and the tests looking for data changes aren't really reliable. 
When we add a full regression test, that should help.

fv3gfs-physics/fv3gfs/physics/physics_state.py
```diff
@@ -1,120 +1,261 @@
...
+    )
+    omga: FloatField = field(
+        metadata={
+            "name": "vertical_pressure_velocity",
+            "units": "Pa/s",
+            "intent": "inout",
+            "full_model_var": True,
+        }
+    )
+    qvapor_t1: FloatField = field(
```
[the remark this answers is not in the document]
rheacangeo: I changed all the '<var>_t1' variables to be named 'physics_updated_<var>'.

driver/pace/driver/driver.py
```diff
@@ -0,0 +1,167 @@
...
+            phis=self.dycore_state.phis,
+        )
+        if not self.namelist.dycore_only:
+            self.physics = Physics(
+                stencil_factory=self.stencil_factory,
+                grid_data=self.grid_data,
+                namelist=self.namelist,
+            )
+            self.physics_state = self.init_physics_from_dycore()
+            self.state_updater = UpdateAtmosphereState(
```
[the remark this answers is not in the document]
elynnwu: Sure, let's do that in a separate PR.

elynnwu (approved):
Looks good! Really like the active_packages in physics init. In the future, for the physics packages that can be turned off from namelist, we should assert to make sure they are consistent with what's passed in active_packages.

fv3core/fv3core/_config.py
```diff
@@ -390,7 +390,13 @@ class Namelist:
     # use_ncep_sst: bool
     # blocksize: int
     # chksum_debug: bool
-    # dycore_only: bool
+    """
+    WARNING: dycore_only may not be used in this model
+     the same way it is in the Fortran version, watch for
+     consequences of these inconsistencies, or more closely
+     parallel the Fortran structure
```
mcgibbon, line 397: Nit: indentation, and allcaps warning is a little more alarm than I think is necessary for this. You could switch to a lowercase "warning" if you think "note" is not enough alarm.
```suggestion
    note: dycore_only may not be used in this model
       the same way it is in the Fortran version, watch for
       consequences of these inconsistencies, or more closely
       parallel the Fortran structure
```

fv3gfs-physics/fv3gfs/physics/physics_state.py
mcgibbon: OK thanks, I didn't realize this only provides an arg to post_init. Very cool! We could maybe follow this same pattern in DycoreState for initializing temporary quantities/storages.

Is there ever a case where the user would want to set `tendency` in MicrophysicsState themselves? Are there any other storages here that are only temporaries/outputs (no input is read) that could also be initialized using the quantity_factory (possibly in another PR)?

fv3gfs-physics/fv3gfs/physics/physics_state.py
```diff
@@ -1,135 +1,211 @@
...
+            "name": "layer_mean_vertical_velocity_microph",
+            "units": "m/s",
+            "intent": "inout",
+        }
+    )
+    prsi: FloatField = field(
+        metadata={"name": "interface_pressure", "units": "Pa", "intent": "inout"}
+    )
+    prsik: FloatField = field(
+        metadata={"name": "log interface_pressure", "units": "Pa", "intent": "inout"}
```
mcgibbon, line 166:
```suggestion
        metadata={"name": "log_interface_pressure", "units": "Pa", "intent": "inout"}
```

fv3gfs-physics/fv3gfs/physics/physics_state.py
```diff
@@ -1,135 +1,211 @@
...
+    )
+    physics_updated_va: FloatField = field(
+        metadata={"name": "physics_northward_wind", "units": "m/s", "intent": "inout"}
+    )
+    delprsi: FloatField = field(
+        metadata={
+            "name": "model_level_pressure_thickness_in_physics",
+            "units": "Pa",
+            "intent": "inout",
+        }
```
mcgibbon, line 136: Very nice metadata!

fv3gfs-physics/fv3gfs/physics/stencils/physics.py
```diff
@@ -169,18 +167,15 @@ def __init__(
         stencil_factory: StencilFactory,
         grid_data: GridData,
         namelist,
-        comm: pace.util.CubedSphereCommunicator,
-        partitioner: TilePartitioner,
-        rank: int,
-        grid_info,
+        active_packages: List[str],
```
mcgibbon, line 170: Can you document what options are available for active_packages, either in a docstring or through a `List[Literal["microphysics", "another_package_name", ...]]` type hint?

pace-util/pace/util/grid/generation.py
```diff
@@ -2217,3 +2217,21 @@ def _reduce_global_area_minmaxes(self):
             self._da_max = max_area
             self._da_min_c = min_area_c
             self._da_max_c = max_area_c
+
+    def split_cartesian_into_storages(self, var: fv3util.Quantity):
+        """
+        Provided a quantity of dims [X_DIM, Y_DIM, CARTESIAN_DIM]
+             or [X_INTERFACE_DIM, Y_INTERFACE_DIM, CARTESIAN_DIM]
+        Split it into separate 2D storages for each cartesian
+        dimension, and return these in a list.
```
mcgibbon, line 2226:
```suggestion
        Provided a quantity of dimensions [X_DIM, Y_DIM, CARTESIAN_DIM]
        or [X_INTERFACE_DIM, Y_INTERFACE_DIM, CARTESIAN_DIM],
        split it into separate 2D storages for each Cartesian
        dimension, and return these in a list.
```

stencils/pace/stencils/testing/grid.py
```diff
@@ -1089,3 +1089,110 @@ def _grid_indexing_axis_offsets(
...
+    Terms used to Apply Physics changes to the Dycore.
+    Attributes:
+      vlon1: x-component of unit lon vector in eastward longitude direction
+      vlon2: y-component of unit lon vector in eastward longitude direction
+      vlon3: z-component of unit lon vector in eastward longitude direction
+      vlat1: x-component of unit lat vector in northward latitude direction
+      vlat2: y-component of unit lat vector in northward latitude direction
+      vlat3: z-component of unit lat vector in northward latitude direction
+      edge_vect_w: factor to interpolate A to C grids at the western grid edge
+      edge_vect_e: factor to interpolate A to C grids at the easter grid edge
```
mcgibbon, line 1106:
```suggestion
      edge_vect_e: factor to interpolate A to C grids at the eastern grid edge
```

stencils/pace/stencils/testing/grid.py
```diff
@@ -1089,3 +1089,93 @@ def _grid_indexing_axis_offsets(
...
+    edge_vect_w: FloatFieldI
+    edge_vect_e: FloatFieldI
+    edge_vect_s: FloatFieldIJ
+    edge_vect_n: FloatFieldIJ
+    es1_1: FloatFieldIJ
+    es1_2: FloatFieldIJ
+    es1_3: FloatFieldIJ
+    ew2_1: FloatFieldIJ
+    ew2_2: FloatFieldIJ
+    ew2_3: FloatFieldIJ
```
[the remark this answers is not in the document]
mcgibbon: Thanks for updating!

Ideally we don't - the way to avoid duplicating it is to have sub-containers as attributes on higher containers, so that the higher containers can say "I have this sub-container attribute" and make use of its documentation. This is what we do in our init calls - the init signatures say "I want this container", and you'd look at the container's documentation for variable descriptions.

When an attribute is actually needed on multiple containers to be read directly on that container, we need to document each use.

mcgibbon (approved):
Looks great! Just a couple documentation typos/suggestions, only one thing missing for docs (a list of active_packages options). Thanks for the fast updates!

PR #82, opened 2021-12-21
Feature/physics cache plan

elynnwu:
## Purpose

Add support for generating and fetching gt_cache for physics performance backend. 

Remove the sections below which do not apply.

## Code changes:

- `pace_physics_cache_setup.sh` includes all jenkins logic
- Add fetch/generate_cachces at top level jenkins similar to fv3core
- Relax override threshold for gpu backend
    - previously, `gtc:gt:gpu` was not tested and did not pass validation
    - In this PR, microphysics max error needed to be relaxed to pass the test

## Infrastructure changes:

- `pace-physics-cache-setup` runs nightly on [jenkins](https://jenkins.ginko.ch/view/pace/job/pace-physics-cache-setup/)


elynnwu:
launch jenkins

.jenkins/actions/fetch_caches.sh
```diff
@@ -0,0 +1,24 @@
...
+        version_file=${CACHE_DIR}/${EXPNAME}/${SANITIZED_BACKEND}/GT4PY_VERSION.txt
+        if [ -f ${version_file} ]; then
+            version=`cat ${version_file}`
+        else
+            version=""
+        fi
+        if [ "$version" == "$GT4PY_VERSION" ]; then
+            if [ -d ${CACHE_DIR}/${EXPNAME}/${SANITIZED_BACKEND}/.gt_cache  ]; then
+                cp -r ${CACHE_DIR}/${EXPNAME}/${SANITIZED_BACKEND}/.gt_cache fv3gfs-physics/.
+                find fv3gfs-physics/. -name m_\*.py -exec sed -i "s|\/scratch\/snx3000\/olifu\/jenkins_submit\/workspace\/pace-physics-cache-setup\/backend\/${SANITIZED_BACKEND}\/experiment\/${EXPNAME}\/slave\/daint_submit/fv3gfs-physics|$(pwd)/fv3gfs-physics|g" {} +
```
mcgibbon, line 17: If you run the existing fetch_caches.sh script inside the fv3gfs-physics directory, it should behave the same as this edited copy. You may have some kind of error with the find command above, and that error might or might not cause a test failure, so be sure to fix that line as well, somehow. I don't really know what it does.

.jenkins/generate_caches.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Jenkins action to run a benchmark of dynamics.py on Piz Daint
+# 3/11/2021, Tobias Wicky, Vulcan Inc
+
+# Syntax:
+# .jenkins/action/run_standlone.sh <option>
+
+## Arguments:
+# $1: <option> which can be either empty, "profile" or "build_cache"
```
mcgibbon, line 10: optional: May want to update the header of the existing fetch/generate caches scripts, while you're in there.

.jenkins/generate_caches.sh
```diff
@@ -0,0 +1,42 @@
...
+backend=$1
+experiment=$2
+SANITIZED_BACKEND=`echo $backend | sed 's/:/_/g'` #sanitize the backend from any ':'
+CACHE_DIR="/scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/${experiment}/${SANITIZED_BACKEND}"
+
+
+test -n "${experiment}" || exitError 1001 ${LINENO} "experiment is not defined"
+test -n "${SANITIZED_BACKEND}" || exitError 1002 ${LINENO} "backend is not defined"
+# caches are built inside fv3gfs-physics
+cd fv3gfs-physics
```
mcgibbon, line 33: If you run the old script inside fv3gfs-physics, it should behave the same.

.jenkins/actions/fetch_caches.sh
mcgibbon: From our discussion, it probably makes sense to do this find and replace in `generate_caches` to replace the path with a constant placeholder string (whatever you like), so that in this script we can replace the placeholder with the current directory.

Very strange that gt4py caches need to have a string referring to the directory the file is in. This should be possible to get programmatically.

.jenkins/pace_physics_cache_setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [[ "${NODE_NAME}" == *"daint"* ]] ; then source ~/.bashrc ; fi
+set -e
+export LONG_EXECUTION=1
+.jenkins/jenkins.sh run_physics_regression_tests ${backend} ${experiment}
+cd buildenv && git reset --hard && cd ..
```
mcgibbon, line 6: Why is this needed, but not for the rest of the repo? Is it because jenkins.sh leaves buildenv in a dirty state? If so, we may want to fix that (but as a todo Jira task, not right now).

.jenkins/pace_physics_cache_setup.sh
```diff
@@ -0,0 +1,8 @@
...
+.jenkins/jenkins.sh run_physics_parallel_regression_tests ${backend} ${experiment}
+.jenkins/generate_caches.sh ${backend} ${experiment}
```
mcgibbon, line 8: Does it make more sense to have pace_physics_cache_setup.sh call jenkins.sh twice, or to make an action that calls these two other actions and then generate_caches.sh, which is called by jenkins.sh once?

fv3gfs-physics/tests/savepoint/translate/overrides/baroclinic.yaml
```diff
@@ -65,9 +65,10 @@ Microph:
   - backend: gtcuda
     cuda_no_fma: true
   - backend: gtc:gt:gpu
+    max_error: 1.1e-10
```
mcgibbon, line 68: Can you describe in the PR description why this needed to be relaxed, and what it was relaxed from and to (since in this file we can only see the "to")?

mcgibbon (changes requested):

.jenkins/pace_physics_cache_setup.sh
elynnwu: Yes, what is strange is that if instead of calling `pace_physics_cache_setup`, we have the same commands in he jenkins exec shell, then buildenv gets reset. I'll put a note for this.

.jenkins/pace_physics_cache_setup.sh
elynnwu: I did try to make a new action that calls both tests, but was getting the same issue with buildenv (and maybe other folders?) being in a dirty state after the serial test. I think this is fine unless there's a preference to get a new action for running both tests working?

mcgibbon: The downside of this mode is a lot of logic in jenkins.sh will run twice, making the plan take longer. Since this isn't a PR plan, that's probably not an issue this time. I'm more concerned that there's an issue with these plans where you can't run scripts consecutively because buildenv is left in a dirty state. Your design is being dictated by that behavior, instead of by what you think makes the most sense. Ideally our scripts shouldn't modify buildenv, and should copy any files they need to modify.

But that's outside the scope of this PR, you can leave this script as is. Thanks for the explanation!

.jenkins/generate_caches.sh
```diff
@@ -0,0 +1,47 @@
...
+    export GT4PY_VERSION=`git submodule status ${PACE_DIR}/external/gt4py | awk '{print $1;}'`
+    echo "GT4PY_VERSION is ${GT4PY_VERSION}"
+fi
+
+CACHE_FILENAME=${CACHE_DIR}/${GT4PY_VERSION}.tar.gz
+
+test -n "${experiment}" || exitError 1001 ${LINENO} "experiment is not defined"
+test -n "${SANITIZED_BACKEND}" || exitError 1002 ${LINENO} "backend is not defined"
+
+cd fv3gfs-physics
```
mcgibbon, line 37: Can this get run from inside `fv3gfs-physics` so it is consistent with where it is run for dycore caching?

.jenkins/actions/fetch_caches.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
mcgibbon, line 1: Unless we call this directly using `jenkins.sh` (I don't think we do?) this should be moved up to `.jenkins`.

.jenkins/jenkins.sh
```diff
@@ -75,7 +76,9 @@ echo "PYTHON env ${python_env}"
 # If the backend is a GTC backend we fetch the caches
 if [[ $backend != *numpy* ]];then
     echo "Fetching for exisintg gt_caches"
-    . ${JENKINS_DIR}/actions/fetch_caches.sh $backend $experiment
+    cd ${PACE_DIR}/fv3gfs-physics
+    . ${JENKINS_DIR}/fetch_caches.sh $backend $experiment
```
mcgibbon, line 80: pre-existing nit, but is there a reason this is executed with a leading `.`? The impact of this is the script gets run as if it's in the local namespace, i.e. writing any variables defined in fetch_caches.sh to the local namespace. I've had issues with this before overwriting SCRIPT_DIR to be the value from the called file.

On main there are no such references and it looks like this can be deleted.
```suggestion
    ${JENKINS_DIR}/fetch_caches.sh $backend $experiment
```

.jenkins/jenkins.sh
```diff
@@ -75,7 +76,9 @@ echo "PYTHON env ${python_env}"
 # If the backend is a GTC backend we fetch the caches
 if [[ $backend != *numpy* ]];then
     echo "Fetching for exisintg gt_caches"
-    . ${JENKINS_DIR}/actions/fetch_caches.sh $backend $experiment
+    cd ${PACE_DIR}/fv3gfs-physics
+    ${JENKINS_DIR}/fetch_caches.sh $backend $experiment
+    cd ${PACE_DIR}
```
mcgibbon, line 81: Not really something you can fix right now/in this PR, so no changes needed: I want to point out this (pre-existing code) hard-codes that every test whose backend doesn't contain numpy using jenkins.sh should fetch caches in the physics directory. For example, if we moved the pace-util tests to use this jenkins.sh, the backend would be an empty string, and caches would get fetched for an empty string backend.

It would be nice if we could run code to fetch caches only for the tests which require fetching caches. I think doing this would require adding a similar concept to "actions" except that they're scripts which are never submitted to a slurm job/docker container/etc., they always run bare-metal, and basically act as hot-swappable patches of code that you would put in jenkins.sh otherwise.

mcgibbon (approved):
Thank you!

PR #84, opened 2021-12-22
Cleanup pace stencils testing 

elynnwu:
## Purpose

When removing global access, we moved some functions (mostly inside `grid.py` into pace.stencils.testing). In this PR, we  move those functions to more fitting location.

## Code changes:

- Made `helper.py` inside `pace.util.grid` which contains `DampingCoefficients`, `*GridData`, and `quantity_wrap` that were previously inside `grid.py`
- As stated in `grid.py`, delete the `axis_offsets` routine in favor of `grid_indexing.axis_offsets`: replace all references to this

## Checklist


pace-util/pace/util/grid/helper.py
```diff
@@ -0,0 +1,513 @@
...
+            (1, shape[1]),
+            axis=1,
+            backend=metric_terms.edge_e.gt4py_backend,
+        )
+        edge_w = utils.make_storage_data(
+            metric_terms.edge_w.data,
+            (1, shape[1]),
+            axis=1,
+            backend=metric_terms.edge_w.gt4py_backend,
+        )
```
mcgibbon, line 177: Can you use `metric_terms.edge_n.storage` here? Why do we need `make_storage_data` from `dsl`? If it is not possible to convert the quantity `metric_terms.edge_w` into a storage using only that object, it feels like there's something wrong with `quantity.storage`.

pace-util/pace/util/grid/helper.py
```diff
@@ -0,0 +1,513 @@
...
+        ak = metric_terms.ak.data
+        bk = metric_terms.bk.data
+        # TODO fix <Quantity>.storage mask for FieldK
+        ak = utils.make_storage_data(
+            ak, ak.shape, len(ak.shape) * (0,), backend=metric_terms.ak.gt4py_backend
+        )
+        bk = utils.make_storage_data(
+            bk, bk.shape, len(bk.shape) * (0,), backend=metric_terms.ak.gt4py_backend
+        )
```
mcgibbon, line 217: Similarly, can we do this without `make_storage_data`?

pace-util/pace/util/grid/helper.py
```diff
@@ -0,0 +1,513 @@
+import dataclasses
+from typing import Any, Optional, Sequence
+
+import pace.util
+from pace.dsl import gt4py_utils as utils
+from pace.dsl.stencil import GridIndexing
+from pace.dsl.typing import FloatFieldI, FloatFieldIJ
```
mcgibbon, line 7: This number of imports from `pace.dsl` are setting off alarm bells for me. Is it possible to make storages in this file without `gt4py_utils`? `quantity_wrap` is only used in one file, `dyn_core.py` in fv3core - can it be put there instead?

The typing is a harder question to answer and we can iterate on after. If we want both util and dsl to be able to depend on gt4py, util only doing so for storages and nothing else, then the type hints for storages will need to live in util and get imported into dsl. However, @jdahm has been telling me FloatField is not technically accurate for storages outside of stencils, so these hints themselves might be overhauled (keeping FloatField hints in dsl for use in stencils, and making a new Storage protocol type hint for use outside of stencils like here).

fv3core/fv3core/stencils/dyn_core.py
```diff
@@ -29,16 +29,10 @@
...
 from pace.dsl.typing import FloatField, FloatFieldIJ, FloatFieldK
-from pace.stencils.testing.grid import (
-    DampingCoefficients,
-    GridData,
-    GridIndexing,
-    axis_offsets,
-    quantity_wrap,
-)
 from pace.util import X_DIM, Y_DIM, Z_DIM, Z_INTERFACE_DIM
+from pace.util.grid import DampingCoefficients, GridData, quantity_wrap
```
mcgibbon, line 35: It appears this is actually the only place we use `quantity_wrap`. The function itself is a pretty private-looking one, that always sets units to "unknown" - does it make sense to put that function directly in this file with a `_` starting its name?

fv3core/tests/savepoint/translate/translate_d_sw.py
```diff
@@ -128,8 +127,8 @@ def __init__(self, grid, namelist, stencil_factory):
         self.out_vars = {"vb": grid.compute_dict_buffer_2d()}
         origin = self.grid.compute_origin()
         domain = self.grid.domain_shape_compute(add=(1, 1, 0))
-        ax_offsets = axis_offsets(self.grid, origin, domain)
         self.stencil_factory = stencil_factory
+        ax_offsets = self.stencil_factory.grid_indexing.axis_offsets(origin, domain)
```
mcgibbon, line 131: Nice! Feels much better to be using this as a method.

mcgibbon (changes requested):

pace-util/pace/util/grid/helper.py
elynnwu: Good point. There was an issue with `edge_e` and `edge_w` having dimension [1, Y_INTERFACE_DIM] with (True, True, False) mask, which kept these from calling `.storage` directly. Added a `UNIT_X_DIM` dimension in Metric Terms to create these fields properly in the first place.

pace-util/pace/util/grid/helper.py
elynnwu: Fixed. Added an optional mask input to Quantity to have mask (True, True, False), then `.storage` works as expected.

pace-util/pace/util/grid/helper.py
elynnwu: Moved `quantity_wrap` to a private function in `dyn_core`. Let's discuss the typing and that can go into the circular dependency PR.

pace-util/pace/util/grid/generation.py
```diff
@@ -84,6 +85,7 @@ def __init__(
                 self.LON_OR_LAT_DIM: 2,
                 self.TILE_DIM: 6,
                 self.CARTESIAN_DIM: 3,
+                self.UNIT_X_DIM: 1,
```
mcgibbon, line 88: Could you use pace.util.X_DIM instead? I don't think there's a need to differentiate UNIT_X_DIM from X_DIM - we similarly don't differentiate X_DIM_COMPUTE from X_DIM_FULL. It's enough that it indicates which axis the dimension is along (which is X_DIM).

pace-util/pace/util/quantity.py
```diff
@@ -259,6 +259,7 @@ def __init__(
         origin: Sequence[int] = None,
         extent: Sequence[int] = None,
         gt4py_backend: Union[str, None] = None,
+        mask: Optional[Tuple[bool, bool, bool]] = None,
```
mcgibbon, line 262: I think we should be able to programmatically generate the mask, without taking it as an input. Something like
```
mask = [
    any(dim in coord_dims for dim in quantity.dims)
    for coord_dims in X_DIMS, Y_DIMS, Z_DIMS
]
```

mcgibbon: You may want to confirm this actually works with someone on the backend (@jdahm?) - my understanding is that mask is specifically for x, y, and z dimensions in that order, always.

mcgibbon: It would be a big help if you could add unit tests to pace-util making sure 2D storages continue to work.

pace-util/pace/util/grid/helper.py
mcgibbon: Doesn't need to be addressed in this PR, but we also might consider putting all the dataclass containers in a file named to describe that that's what's inside it. I'm not sure what to expect from `grid.helper`.

pace-util/pace/util/quantity.py
jdahm: @mcgibbon Yeah, I think we should be able to automatically generate those from a quantity like you wrote.

pace-util/pace/util/grid/generation.py
elynnwu: Changed to `pace.util.X_DIM`

pace-util/pace/util/quantity.py
elynnwu: This is now updated to detect the mask inside Quantity.

pace-util/pace/util/initialization/allocator.py
```diff
@@ -54,29 +55,52 @@ def from_backend(cls, sizer: SubtileGridSizer, backend: str):
...
         origin = self._sizer.get_origin(dims)
         extent = self._sizer.get_extent(dims)
         shape = self._sizer.get_shape(dims)
+        mask = tuple(
+            [
+                any(dim in coord_dims for dim in dims)
+                for coord_dims in [X_DIMS, Y_DIMS, Z_DIMS]
+            ]
+        )
+        spatial_dims = [i for j in [X_DIMS, Y_DIMS, Z_DIMS] for i in j]
```
mcgibbon, line 98: I manually tested the code you've written does in fact get the correct spatial_dims, but I still cannot tell from reading this Python list comprehension how it works. This should be more straightforward:
```suggestion
        spatial_dims = X_DIMS + Y_DIMS + Z_DIMS
```

pace-util/pace/util/initialization/allocator.py
```diff
@@ -54,29 +55,52 @@ def from_backend(cls, sizer: SubtileGridSizer, backend: str):
...
+        extra_dims = [i for i in dims if i not in spatial_dims]
+        if len(extra_dims) > 0 or not dims:
+            mask = None
         try:
-            data = allocator(shape, dtype=dtype, default_origin=origin)
+            data = allocator(shape, dtype=dtype, mask=mask)
```
mcgibbon, line 103: Was the default_origin removed intentionally? 
```suggestion
            data = allocator(shape, dtype=dtype, default_origin=origin, mask=mask)
```

pace-util/pace/util/quantity.py
```diff
@@ -284,6 +284,13 @@ def __init__(
         else:
             extent = tuple(extent)
 
+        mask = tuple(
+            [
+                any(dim in coord_dims for dim in dims)
+                for coord_dims in [constants.X_DIMS, constants.Y_DIMS, constants.Z_DIMS]
+            ]
+        )
```
mcgibbon, line 292: optional nit: You could put this inside the if block for gt4py_backend is not None, so it only gets computed when needed.

pace-util/pace/util/quantity.py
```diff
@@ -392,12 +399,19 @@ def storage(self):
...
         return self._storage
 
-    def _initialize_storage(self, data, origin, gt4py_backend: str):
+    def _initialize_storage(self, data, origin, gt4py_backend: str, mask: Tuple):
+        spatial_dims = [
+            i for j in [constants.X_DIMS, constants.Y_DIMS, constants.Z_DIMS] for i in j
+        ]
+        extra_dims = [i for i in self._dims if i not in spatial_dims]
+        if len(extra_dims) > 0 or not self._dims:
+            mask = None
```
mcgibbon, line 408: Instead of passing `dims` silently through an object attribute, can you put this if statement where `mask` is computed so the passed value for `mask` will be None in this case, and then remove the attribute `self._dims`? This should make it clearer when reading the code that computes the mask that we only compute/use the mask when there are no non-spatial dimensions.

pace-util/pace/util/quantity.py
```diff
@@ -392,12 +399,19 @@ def storage(self):
             )
         return self._storage
 
-    def _initialize_storage(self, data, origin, gt4py_backend: str):
+    def _initialize_storage(self, data, origin, gt4py_backend: str, mask: Tuple):
+        spatial_dims = [
+            i for j in [constants.X_DIMS, constants.Y_DIMS, constants.Z_DIMS] for i in j
+        ]
+        extra_dims = [i for i in self._dims if i not in spatial_dims]
```
mcgibbon, line 406: This code to compute extra_dims is duplicated. Instead of pulling it into a function, can you add a `pace.util.SPATIAL_DIMS` constant equal to `X_DIMS + Y_DIMS + Z_DIMS`, defined in the same file where these constants are defined? Then the one-liner for extra_dims is simple and fine to duplicate.

mcgibbon (commented):
Some small change comments for the updates to quantity/allocator, let me know when you've updated and it should be a quick final approve.

fv3gfs-physics/tests/savepoint/translate/translate_fv_update_phys.py
```diff
@@ -6,8 +6,8 @@
 import pace.util
 from pace.dsl.typing import FloatField, FloatFieldIJ
 from pace.stencils.fv_update_phys import ApplyPhysics2Dycore
-from pace.stencils.testing.grid import DriverGridData
 from pace.stencils.testing.parallel_translate import ParallelTranslate2Py
+from pace.util.grid import DriverGridData
```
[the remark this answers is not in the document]
elynnwu: Yes, those need to be moved outside physics and into driver.

pace-util/pace/util/initialization/allocator.py
```diff
@@ -54,29 +55,52 @@ def from_backend(cls, sizer: SubtileGridSizer, backend: str):
...
+    ):
         return self._allocate(self._numpy.zeros, dims, units, dtype)
 
-    def ones(self, dims: Sequence[str], units: str, dtype: type = float):
+    def ones(
+        self,
+        dims: Sequence[str],
+        units: str,
+        dtype: type = float,
+    ):
```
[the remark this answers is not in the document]
elynnwu: Good question, I had originally added an extra argument for mask, that's when the reformat happened. I'm not sure why it didn't get reformatted back, very odd. Maybe it was okay either way for this length...

pace-util/pace/util/initialization/allocator.py
elynnwu: Great catch, it should not have been deleted.

pace-util/pace/util/quantity.py
elynnwu: Good idea, done!

pace-util/pace/util/quantity.py
elynnwu: Added `pace.util.SPATIAL_DIMS`

pace-util/pace/util/quantity.py
```diff
@@ -306,8 +307,22 @@ def __init__(
...
+                        for coord_dims in [
+                            constants.X_DIMS,
+                            constants.Y_DIMS,
+                            constants.Z_DIMS,
+                        ]
+                    ]
+                )
             self._storage, self._data = self._initialize_storage(
-                data, origin, gt4py_backend
+                data, origin, gt4py_backend, mask
```
mcgibbon, line 325: nit: just a little worried about passing as positional args when we get to more than 3 arguments.
```suggestion
                data, origin=origin, gt4py_backend=gt4py_backend, mask=mask
```

pace-util/pace/util/quantity.py
```diff
@@ -284,6 +284,7 @@ def __init__(
         else:
             extent = tuple(extent)
 
+        self._dims = dims
         if isinstance(data, (int, float, list)):
             data = np.asarray(data)
         elif gt4py is not None and isinstance(data, gt4py.storage.storage.Storage):
```
mcgibbon, line 290: I might suggest not merging this, but pointing it out so you're aware - there's a bug here where if someone passes a list of floats to initialize the quantity with a z-dimension, and includes a gt4py backend, the storage attribute won't work when you'd expect it should. Since we never initialize quantities this way (perhaps the list option should be removed altogether), this is not an issue.

If you merge this suggestion it might cause unforeseen test failures, so it's really up to you.

```suggestion
        if gt4py is not None and isinstance(data, gt4py.storage.storage.Storage):
```

mcgibbon (approved):
Thanks! Really nice! Two code suggestions to merge and one to probably ignore, then LGTM.

PR #86, opened 2021-12-22
driver final review adjustments

rheacangeo:
## Purpose

Making small fixes that were requested in #79, but didn't happen because automerge was on. 

## Code changes:

- make an explicit list of python packages allowed as a type-hint
- change WARNING to note
- fix some typos

### DIFF
--- a/driver/pace/driver/run.py
+++ b/driver/pace/driver/run.py
@@ -1,12 +1,13 @@
 import os
 import sys
 from dataclasses import fields
-from typing import List, Literal, Optional, Tuple
+from typing import List, Optional, Tuple
 
 import click
 import f90nml
 import numpy as np
 from mpi4py import MPI
+from typing_extensions import Literal
 
 import fv3core._config
 import fv3core.initialization.baroclinic as baroclinic_init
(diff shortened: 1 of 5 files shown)


driver/pace/driver/run.py
```diff
@@ -1,12 +1,13 @@
...
 import sys
 from dataclasses import fields
-from typing import List, Literal, Optional, Tuple
+from typing import List, Optional, Tuple
 
 import click
 import f90nml
 import numpy as np
 from mpi4py import MPI
+from typing_extensions import Literal
```
mcgibbon, line 10: nit: It's only necessary to import this from typing_extensions if we're on python below 3.8, but since this depends on gt4py and gt4py depends on python >=3.8 we should be able to import it from `typing`. I don't mind either way but wanted to make sure you know this isn't necessary for the versions we use.

fv3gfs-physics/fv3gfs/physics/stencils/physics.py
```diff
@@ -2,6 +2,7 @@
 
 import gt4py.gtscript as gtscript
 from gt4py.gtscript import BACKWARD, FORWARD, PARALLEL, computation, exp, interval, log
+from typing_extensions import Literal
```
mcgibbon, line 5: Same nit about import location.

mcgibbon (approved):
LGTM! Minor comment about where we import Literal from.

driver/pace/driver/run.py
rheacangeo: ah it was complaining at me because I'm running the linter using python 3.7 then. we import it from typing_extensions in another part of the code...

mcgibbon: if you want to leave it to get it in for now that’s fine, there isn’t really any harm in it.

PR #96, opened 2022-01-12
driver can run from serialized init data again and is tested

rheacangeo:
## Purpose

With the Driver we added the option to initialize from serialized data, because there are experiments we want to run that don't start from the baroclinic analytic state init. We didn't add a test for this functionality, however, and a recent change broke it. This fixes it and adds a driver test that just checks run.py can run, does not do regressions yet (coming soon) 

## Code changes:

- update method in the fv3core/testing/translate_fvdynamics code that makes a DycoreState from serialized data
- update MAX_SIZE for future stencil to accommodate dycore+physics+driver stencils
- synchronization of the state before writing out is fixed to synchronize the storages now that the state fields are storages

## Infrastructure changes:

- added test_serialized_init to the driver Makefile that checks the driver/run.py script does not crash when starting from serialized data 

## Checklist


dsl/pace/dsl/future_stencil.py
```diff
@@ -50,7 +50,7 @@ class StencilTable(object, metaclass=Singleton):
 
     DONE_STATE: int = -1
     NONE_STATE: int = -2
-    MAX_SIZE: int = 200
+    MAX_SIZE: int = 220
```
elynnwu, line 53: Only 20 more, physics is so simple! Will try to remember this as we add more stencils in physics.

elynnwu (approved):
lgtm!

dsl/pace/dsl/future_stencil.py
rheacangeo: very true! the actual size is 2*MAX_SIZE + 1, so increased from 401 to 441. Yeah definitely, we can change this number as/when we need, and also the distributed compilation code may end up changing and not using this

PR #99, opened 2022-01-18
Driver demo for AMS

mcgibbon:
## Purpose

Update `pace.driver.run` for presentation at AMS.

This PR updates `pace.driver.run` to be an example for presentation at AMS, using baroclinic initialization and yaml configuration.

## Code changes:

- increased StencilTable.MAX_SIZE to 800
- removed unused cupy import in pace.dsl.stencil
- refactored StencilConfig to a dataclass which can be initialized using dacite
- imported DycoreState to top level of fv3core
- corrected type hint on consv_te/conserve_total_energy in configuration classes and function arguments
- removed unused arguments `lat_agrid` and `adiabatic` from baroclinic initialization code
- fixed `init_zeros` method of `PhysicsState`
- refactored timestep to be a call-time argument on Microphysics, Physics
- added CubedSphereCommunicator.from_layout constructor method
- added error in CubedSphereCommunicator init if given a communicator with a size not equal to the total ranks of the given partitioner
- ignoring divide-by-zero warnings in grid generation
- added DycoreToPhysics class to pace.stencils.update_atmos_state, which copies data from the dycore to the physics
- imported DycoreToPhysics and UpdateAtmosphereState to top level of `pace.stencils`
- added support for built-in `datetime` to `ZarrMonitor`
- moved `init_dycore_state_from_serialized_data` from `translate_fvdynamics.py` to the driver runfile

## Requirements changes:

- added dacite as requirement for driver, to initialize from yaml configuration

## Infrastructure changes:

- Added READMEs for driver and driver/examples
- tox is now called with `--recreate` flag, to ensure its test environment is up to date
- added test_driver workflow for CircleCI
- added `.jenkins/test_driver.sh` script to run driver tests inside Docker
- removed docker and data logic from `driver/Makefile`
- added `driver/requirements.txt`
- driver now depends on dacite for yaml data loading
- test_driver now mocks calls to stencil-running objects like Physics, Dynamics, and coupling classes
- added sanity checks to driver tests for the number of times components are called
- fixed use of total_ranks fixture in several pace-util tests involving communication

## Checklist


mcgibbon:
launch jenkins

driver/pace/driver/run.py
```diff
@@ -1,292 +1,379 @@
...
-        elif dycore_init_mode == "baroclinic":
-            # create an initial state from the Jablonowski & Williamson Baroclinic
-            # test case perturbation. JRMS2006
+    def store(self, time: datetime, state: DriverState):
+        zarr_state = {"time": time}
+        for name in self.config.names:
+            try:
+                quantity = getattr(state.dycore_state, name)
+            except AttributeError:
+                quantity = getattr(state.physics_state, name)
```
elynnwu, line 137: Not right now - have you thought about how we should handle diagnostics variables inside physics?

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -2235,7 +2227,21 @@ def setupm(self):
...
+        # Small time step
+        self._dts = timestep / self._ntimes
+        self._dt_rain = self._dts * 0.5
+        self._rdts = 1.0 / self._dts
+        self._dt_evap = 0.5 * self._dts if self.namelist.fast_sat_adj else self._dts
+        self._fac_i2s = 1.0 - np.exp(-self._dts / self.namelist.tau_i2s)
+        self._fac_g2v = 1.0 - np.exp(-self._dts / self.namelist.tau_g2v)
+        self._fac_v2g = 1.0 - np.exp(-self._dts / self.namelist.tau_v2g)
+        self._fac_imlt = 1.0 - np.exp(-0.5 * self._dts / self.namelist.tau_imlt)
+        self._fac_l2v = 1.0 - np.exp(-self._dt_evap / self.namelist.tau_l2v)
```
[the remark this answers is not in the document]
elynnwu: Oh this is because time of day?

driver/pace/driver/run.py
```diff
@@ -1,292 +1,379 @@
...
+            n_split=self.config.dycore_config.n_split,
+            do_adiabatic_init=False,
+            timestep=float(timestep),
         )
 
-    def step_physics(self):
-        self.physics(self.physics_state)
-        self.state_updater(self.dycore_state, self.physics_state)
+    def _step_physics(self, timestep: float):
+        self.dycore_to_physics(
```
elynnwu, line 322: Thanks for adding this!

driver/pace/driver/run.py
```diff
@@ -1,292 +1,379 @@
...
+            communicator=communicator,
+            quantity_factory=quantity_factory,
         )
-        return quantity_factory, stencil_factory
+        raise NotImplementedError()
+
+
+@dataclasses.dataclass(frozen=True)
+class DiagnosticsConfig:
+    path: str
```
elynnwu, line 114: We originally had `Path(MODEL_OUT_DIR).mkdir(parents=True, exist_ok=True)` to make output folder. For here, do you have to make a directory first?

elynnwu: Also, not for this PR, but would reading Diagnostics be a function in this class? Right now we have some limited plotting capabilities which read npy files.

elynnwu (commented):
Overall, really like the design for driver. Thanks for cleaning up many loose ends! 
My only concern is that without the ability to initialize from either serialized data or restart files, we cannot easily write a savepoint test for Driver-In/Out for validation. Also, it means that I still cannot move away from that dyn_phy script for running longer time simulation since I need to have it restart from baroclinic day 10. From standup, we agreed that this doesn't need to be in this PR.

mcgibbon:
> we cannot easily write a savepoint test for Driver-In/Out for validation

Generally we don't execute runfiles for savepoint tests - if we add savepoint tests for driver, I imagine we will want to have the "initialize model state" code in the test infrastructure, the same way it is for our other packages.

For wanting to initialize from partway through a baroclinic run, I'd like to add a Jira task to add to/from disk methods to DriverState and possibly also DynamicsState and PhysicsState, and infrastructure in pace.driver.run to write checkpoints periodically to disk, and to initialize from those checkpoints. Does this sound like it fits your use case?

driver/pace/driver/run.py
mcgibbon: You don't have to make the directory itself, though you do have to make the directory you're going to put it in. This is probably safer behavior, because of the case where someone has a typo in a parent directory.

Ideally, the saved data should be readable by an external package, and we shouldn't need a custom object. For the one I wrote, the stored dataset is a zarr you can open with xarray. Do you mean something else by "reading Diagnostics"?

driver/pace/driver/run.py
mcgibbon: From the list of "would be nice to do" things from Lucas, it sounds like they'd like to be able to more easily add "deep diagnostics" in a somewhat performant way. The pattern I wrote here doesn't support that, but was easy to write for this example.

In the longer term, we probably want to design an API for registering (saying "hey I can compute this if you want it" from deep code), selecting (user configures which of those they actually want) and storing (when the code runs, the diagnostics need to be computed/saved) diagnostics.

We'd probably want to propagate a class through all of the init routines to do this, but there are other ways. One alternative is to have a context manager which sets the global diagnostic gatherer, like
```
with diagnostic_manager_1:
    driver_1.step()
with diagnostic_manager_2:
    driver_2.step()
```
But this would be a little less safe and a little more confusing.

driver/pace/driver/run.py
```diff
@@ -1,292 +1,379 @@
...
-
-    for t in range(int(time_steps)):
-        driver.step(
-            do_adiabatic_init=do_adiabatic_init,
-            timestep=bdt,
-        )
-
-        if t % 5 == 0:
-            comm.Barrier()
+def cli(config_path: str):
```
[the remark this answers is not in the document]
mcgibbon: command-line interface. I'll rename this to `command_line`.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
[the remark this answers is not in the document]
elynnwu: I'm concerned about having python code in runtime. Also, none of these variables actually change with time right? So why would we want them in call time?

driver/pace/driver/run.py
elynnwu: Thanks for the response. We should have this in our "would be nice to do" list.

driver/pace/driver/run.py
elynnwu: Got it. We had it so that it will mkdir -p a model_output folder, but I can see that it could be bad if someone passes the wrong absolute path then a folder gets made somewhere else. So this is fine.

Oh good, I didn't realize it's as easy as just opening from xarray. No need to have a read functionality.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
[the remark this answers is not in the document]
mcgibbon: I also don't like having the python code in runtime, I think ideally these should be computed inside stencils. For now I've moved these to a `set_timestep` routine which should only call these compute functions once the first time it's called, and use pre-computed values on subsequent steps.

The way we run the model in our driver, timestep does not change with time. However there are many reasons you might want to initialize a single Microphysics object and call it with different timesteps. For example, to assess the sensitivity of predictions to timestep, or to run a variable-timestep model.

driver/examples/plot_output.py
```diff
@@ -0,0 +1,31 @@
+import xarray as xr
+import zarr
+
+
+try:
+    import matplotlib.pyplot as plt
```
elynnwu, line 6: Maybe you can put this install in `driver/examples/create_venv.sh`?

elynnwu: or is the creat_venv script only supposed to run inside docker? either way, this is optional. I think most python user would know how to maplotlib anyways.

driver/examples/zarr_to_nc.py
```diff
@@ -0,0 +1,15 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Converts zarr directory stores to netcdf"
+    )
+    parser.add_argument("zarr_in", type=str, help="path of zarr to convert")
+    parser.add_argument("netcdf_out", type=str, help="output netcdf")
+    args = parser.parse_args()
+    ds: xr.Dataset = xr.open_zarr(store=zarr.DirectoryStore(args.zarr_in))
+    ds.to_netcdf(args.netcdf_out)
```
elynnwu, line 15: Can you add an example command in README for this script?

driver/pace/driver/run.py
```diff
@@ -185,108 +183,288 @@ def _init_physics_state_from_dycore_state(
...
                     field.metadata["units"],
                     dtype=float,
                 )
-        return PhysicsState(
+        return fv3gfs.physics.PhysicsState(
             **initial_storages,
-            quantity_factory=self._quantity_factory,
-            active_packages=physics_packages,
+            quantity_factory=quantity_factory,
+            active_packages=["microphysics"],
```
elynnwu, line 196: Can you revert to `active_packages=physics_packages` so that we don't forget?

elynnwu: Also `physics_packages` should be an input to this function

driver/pace/driver/run.py
```diff
@@ -1,181 +1,179 @@
...
-import f90nml
-import numpy as np
+import dacite
+import yaml
+import zarr.storage
 from mpi4py import MPI
-from typing_extensions import Literal
 
+import fv3core
 import fv3core.initialization.baroclinic as baroclinic_init
```
elynnwu, line 14: [not for this PR]: but we should move this function to driver too right?

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -11,6 +11,8 @@
 from pace.dsl.typing import Float, FloatField, FloatFieldIJ, Int
 from pace.util.grid import GridData
 
+from .._config import PhysicsConfig
```
elynnwu, line 14: I think you put it in the top level too right? Can we do `from fv3gfs.physics import PhysicsConfig`? The two dots are messing with my brain

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -2234,8 +2226,31 @@ def setupm(self):
         self._csmlt = csmlt
         self._cgmlt = cgmlt
         self._ces0 = constants.EPS * es0
+        self._set_timestep(-1.0)
```
elynnwu, line 2229: is it -1 by default so we always take mp_time?

elynnwu: if that's the case, maybe we should change it to have:
`self._mpdt = self.namelist.mp_time`?

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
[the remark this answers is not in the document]
elynnwu: Yes, I agree. I think what you have it now is fine for this PR. We can think about how we want to handle this later.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
```diff
@@ -2234,8 +2226,31 @@ def setupm(self):
...
+        self._fac_i2s = 1.0 - np.exp(-self._dts / self.namelist.tau_i2s)
+        self._fac_g2v = 1.0 - np.exp(-self._dts / self.namelist.tau_g2v)
+        self._fac_v2g = 1.0 - np.exp(-self._dts / self.namelist.tau_v2g)
+        self._fac_imlt = 1.0 - np.exp(-0.5 * self._dts / self.namelist.tau_imlt)
+        self._fac_l2v = 1.0 - np.exp(-self._dt_evap / self.namelist.tau_l2v)
+        self._timestep = timestep
 
-    def __call__(self, state: MicrophysicsState):
+    def __call__(self, state: MicrophysicsState, timestep: float):
+        self._set_timestep(timestep)
```
elynnwu, line 2253: This is not supposed to be here right? or is some logic missing here? I see that you have another function called `_update_timestep_if_needed` but I don't see it used.

fv3gfs-physics/fv3gfs/physics/stencils/physics.py
```diff
@@ -14,6 +14,8 @@
 from pace.dsl.typing import Float, FloatField
 from pace.util.grid import GridData
 
+from .._config import PhysicsConfig
```
elynnwu, line 17: `from fv3gfs.physics import PhysicsConfig`

elynnwu (approved):
Looks good! Some minor comments.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
mcgibbon: Thanks, fixed it to call `update_timestep_if_needed`.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
mcgibbon: It's -1 by default so that it gets updated on the first timestep, and so that if we have a bug where we don't update the first time this is called then the behavior is weird enough that we should notice.

fv3gfs-physics/fv3gfs/physics/stencils/microphysics.py
mcgibbon: The short version is, within a package we need to use relative imports - top level imports are for "other" packages (e.g. in physics, for stencils/dsl/util).

Within the `physics` package, we can't import everything from `physics` top level, because this would cause a variety of circular dependencies. In this case, `physics.__init__` imports `physics.stencils.microphysics.Microphysics` which would in turn import `physics.__init__` causing a circular dependency.

driver/pace/driver/run.py
mcgibbon: Possibly, possibly not. I believe this function is unit tested in fv3core? There are some reasons to want to be able to initialize a DycoreState even if you don't have the `driver` package. It doesn't depend on any non-dycore packages, and conceptually it's very coupled to `DycoreState`, so it's probably fine for it to live in fv3core.

If this function also initialized physics state, it would need to be in `driver`.

driver/examples/zarr_to_nc.py
mcgibbon: Added.

driver/examples/plot_output.py
mcgibbon: I wanted to leave matplotlib off since it's not a "requirement" of the driver. Hopefully between the line I just added to the instructions and this error message, it should be easy enough to figure out.

driver/examples/README.md
```diff
@@ -0,0 +1,38 @@
...
+In the future, we will add examples for the compiled backends we support.
+
+Note that on the baroclinic test case example you will see significant grid imprinting in the first hour time evolution.
+Rest assured that this duplicates the behavior of the original Fortran code.
+
+## Docker
+
+To run a baroclinic c12 case with Docker in a single command, run `run_docker.sh`.
+This example will start from the Python 3.8 docker image, install extra dependencies and Python packages, and execute the example, leaving the output in this directory.
+There is also a script `plot_output.py` which will show the
```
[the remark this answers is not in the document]
mcgibbon: Thanks, I fixed it.

fv3gfs-physics/fv3gfs/physics/stencils/physics.py
mcgibbon: Leaving as-is since this is an import from physics to physics.

mcgibbon:
launch jenkins

mcgibbon:
All tests passed except for one savepoint test which failed due to a g++ crash, probably due to my aborting the cache build this morning. Going to merge regardless.

PR #103, opened 2022-01-28
Feature/driver init from serialbox

elynnwu:
## Purpose

In this PR, we add the capability of initializing from serialbox in driver. The capability of using serialized grid is also added.

## Code changes:

- Dockerfile: include serialbox in PYTHONPATH to avoid having to append path
- driver/run.py: 
    - `DriverState` now takes in `grid_data`, `damping_coefficients`, and `driver_grid_data`
    - `SerialboxConfig` implemented: an example yaml is included, where a namelist override can be passed in `dycore_config` and `physics_config`. It is also necessary to pass in nx_tile, nz, dt_atmos, layout, and time information in the yaml file. 
- Move `TranslateUpdateDWindsPhys` to stencils so it can be imported in driver
- pace/util/namelist: Allow date/time information from fortran namelist so that `SerialboxConfig` can use it
- translate_physics: `edge_vect_s/n` mask updated

## Infrastructure changes:

- test_driver: a new driver test that initializes from serialbox, this runs as part of `test_driver` on jenkins
    - note that we currently use `fv3gfs-physics` default data but could switch to driver specific data once that's available 

## Checklist


driver/pace/driver/run.py
```diff
@@ -116,82 +130,179 @@ class SerialboxConfig(InitializationConfig):
...
+        )
+        stencil_factory = StencilFactory(
+            config=self.stencil_config, grid_indexing=grid.grid_indexing
+        )
+        driver_grid_info_object = TranslateUpdateDWindsPhys(
+            grid, self.namelist, stencil_factory
+        )
+        extra_grid_data = driver_grid_info_object.collect_input_data(
+            ser, ser.get_savepoint("FVUpdatePhys-In")[0]
+        )
```
mcgibbon, line 198: It's unfortunate that we need a stencil factory to do something that involves creating no stencils, but it looks like that is in fact the case because the translate infrastructure uses low-level configuration from StencilFactory in combination with `make_storage_from_shape` instead of using `QuantityFactory`.

I think though that we only need the `backend` value. You should be able to remove the other configuration options above (rebuild, validate_args, format_source, device_sync) and provide a nonsensical value for them (None, probably) that would cause errors if we ever did create stencils through this pathway. This would reduce the dataclass API above to only what's needed.

driver/pace/driver/run.py
```diff
@@ -116,82 +130,179 @@ class SerialboxConfig(InitializationConfig):
...
+        import serialbox
+
+        serializer = serialbox.Serializer(
+            serialbox.OpenModeKind.Read,
+            self.path,
+            "Generator_rank" + str(communicator.rank),
+        )
+        return serializer
+
+    def get_grid_data_damping_coeff_and_driver_grid(
```
mcgibbon, line 229: Can these be private methods?

driver/pace/driver/run.py
```diff
@@ -116,82 +130,179 @@ class SerialboxConfig(InitializationConfig):
...
+        savepoint_in = ser.get_savepoint("FVDynamics-In")[0]
+        stencil_factory = StencilFactory(
+            config=self.stencil_config, grid_indexing=grid.grid_indexing
+        )
+        translate_object = TranslateFVDynamics([grid], self.namelist, stencil_factory)
+        input_data = translate_object.collect_input_data(ser, savepoint_in)
+        dycore_state = translate_object.state_from_inputs(input_data)
+        return dycore_state
+
+    def _init_physics_state(
```
mcgibbon, line 301: You can delete this method and call the code below directly where you'd call this function, since the call signatures are similar size and the code contained here would be more explicit.

driver/pace/driver/run.py
```diff
@@ -116,82 +130,179 @@ class SerialboxConfig(InitializationConfig):
...
+
+    @property
+    def stencil_config(self) -> pace.dsl.stencil.StencilConfig:
+        return pace.dsl.stencil.StencilConfig(
+            backend=self.backend,
+            rebuild=self.rebuild,
+            validate_args=self.validate_args,
+        )
+
+    def get_serialized_grid_damping_coeff_and_driver_grid(
```
mcgibbon, line 160: Can these be private methods/attributes? They should only be called from inside this class.

driver/pace/driver/run.py
```diff
@@ -279,11 +390,42 @@ def total_time(self) -> timedelta:
...
+        return kwargs
+
     @classmethod
     def from_dict(cls, kwargs: Dict[str, Any]) -> "DriverConfig":
-        kwargs["layout"] = tuple(kwargs["layout"])
         initialization_type = kwargs["initialization_type"]
-        if initialization_type == "baroclinic":
+        if initialization_type == "serialbox":
+            config = kwargs.get("initialization_config", {})
+            config.update(kwargs.get("stencil_config", {}))
```
mcgibbon, line 416: Are these lines necessary? My understanding was that dacite does this automatically, but I could be wrong.
```suggestion
```

driver/pace/driver/run.py
```diff
@@ -279,11 +390,42 @@ def total_time(self) -> timedelta:
...
+                data_class=SerialboxConfig,
+                data=config,
+                config=dacite.Config(strict=True),
+            )
+            kwargs = cls._set_kwargs_from_serialbox(
+                kwargs, kwargs["initialization_config"]
+            )
+            return dacite.from_dict(
+                data_class=cls, data=kwargs, config=dacite.Config(strict=True)
+            )
```
mcgibbon, line 427: This return statement here makes it a lot harder to tell what this function does. There's code below that does necessary processing to integrate the top level driver configuration with values in DycoreConfig and PhysicsConfig. Either that code is duplicated somehow in `set_kwargs_from_serialbox`, which makes it harder to maintain that code (has to be updated in two places), or that code is not duplicated and the behavior of the rest of the configuration depends on how the initialization is set.

It also looks like right now the value of `initialization_type` makes the non-initialization part of the driver configuration yaml behave differently in ways I wouldn't expect as a user.

Is there any way this can be reduced to `initialization_class=SerialboxConfig` with all the concerns met by `set_kwargs_from_serialbox` treated automatically within the `SerialboxConfig` class? You have full control over the attributes of SerialboxConfig to make it work when you call `dacite.from_dict`, and you have access to a `__post_init__` dataclass method to add/modify arbitrary attributes afterwards.

mcgibbon (changes requested):

driver/examples/configs/baroclinic_c12_from_serialbox.yaml
```diff
@@ -0,0 +1,23 @@
+stencil_config:
```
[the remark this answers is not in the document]
elynnwu: Do we not want to have this in driver/Makefile at all? I thought that we would want to have this in examples because while serialbox may not be a desired option for users, the restart config is expected to be used commonly. So shouldn't we have all the examples?
I think documentation on how to use this yaml file is a good idea (will add this), not just externally but also nice to see easily what options we support and how to use it.

driver/Makefile
```diff
@@ -5,3 +5,6 @@ test:
 
 test_mpi:
 	$(MPIRUN_CALL) python3 -m mpi4py -m pace.driver.run examples/configs/baroclinic_c12.yaml
+
+test_mpi_serialbox:
```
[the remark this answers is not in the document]
elynnwu: Fair enough. Do we expect users to run from make file? Because there're some instructions in driver/README, and they don't reference the makefile either.

driver/pace/driver/run.py
```diff
@@ -64,6 +71,11 @@ def get_driver_state(
         metric_terms = pace.util.grid.MetricTerms(
             quantity_factory=quantity_factory, communicator=communicator
         )
+        grid_data = pace.util.grid.GridData.new_from_metric_terms(metric_terms)
+        damping_coeffient = DampingCoefficients.new_from_metric_terms(metric_terms)
+        driver_grid_data = pace.util.grid.DriverGridData.new_from_metric_terms(
+            metric_terms
```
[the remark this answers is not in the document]
elynnwu: I think we need to consolidate `GridData` and `DriverGridData`, mostly just rolling whatever in driver grid data to grid data. There're also some additional changes we need to add in fortran grid savepoint, currently the variables in `DriverGridData` are not in that savepoint.

driver/pace/driver/run.py
```diff
@@ -116,82 +130,179 @@ class SerialboxConfig(InitializationConfig):
...
-    npx: int
-    npy: int
-    npz: int
-    layout: List[int]
+    serialized_grid: bool
     backend: str
+    rebuild: bool
+    validate_args: bool
+    format_source: bool
+    device_sync: bool
```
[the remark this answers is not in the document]
elynnwu: Yes this is because we need to make a `stencil_config` to use TranslateUpdateDWindsPhys. When we include these grid variables in the gridinfo savepoint, it should look a bit cleaner.

driver/pace/driver/run.py
elynnwu: Sounds good, keeping `backend` only here. Again, this should look cleaner when we consolidate `GridData` and `DriverGridData`.

elynnwu: Making a jira card for that now

driver/pace/driver/run.py
elynnwu: Changed to private methods

driver/pace/driver/run.py
elynnwu: Changed the last two. The namelists are actually called in `_set_kwargs_from_serialbox` inside `DriverConfig`. We override the options to be from fortran namelist there.

driver/pace/driver/run.py
elynnwu: It is necessary because we're making a SerialboxConfig first before making DriverConfig class, and we need backend from stencil_config.

driver/pace/driver/run.py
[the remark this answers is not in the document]
rheacangeo: I thought the direction we wanted to go was distinct DriverGridData, PhysicsGridData, and GridData(which could be named DycoreGridData)? You'd rather have one large bucket that has all 3 and contains irrelevant variables when running dycore only or physics only?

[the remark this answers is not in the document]
mcgibbon: I read what @elynnwu wrote as being that we can have a `GridData` class with `grid_data.dycore`, `grid_data.physics` and `grid_data.<whatever we call the conversion part currently referred to as DriverGridData>`. Keeps the convenience of one place to initialize or access the data at the top level, but also the nice interfaces at lower levels where you only need to create relevant variables.

I'm more open on this one versus the configuration classes though, since grid data can be generated with very few free parameters/config options. One bin can be OK as long as there are very few free parameters needed to initialize it.

driver/Makefile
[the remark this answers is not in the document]
mcgibbon: I don't expect user to run this file soon - the Makefile is here for our development purposes. test_mpi is useful to iterate locally on code without the overhead of compiling/running docker. test_mpi_serialbox could also be useful for certain development, but there's no way to run it other than using `.jenkins/test_driver.sh` since it explicitly depends on being in a docker environment (specifically, having a test_data folder at the filesystem root).

You could get it running automatically through this make target, but I think test_driver.sh should be enough.

driver/pace/driver/run.py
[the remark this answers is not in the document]
elynnwu: This is a good point. I hadn't thought about the case where a user just wants to run one component when I made the comment. I think we will have to go through the variables one by one to see which way makes more sense, updating the jira card now.

driver/pace/driver/run.py
elynnwu: Okay updated per our discussion. It's still not super clean because I couldn't find a way to do this nicely without adding more inputs in the serialbox config yaml. More comments below.

fv3core/fv3core/_config.py
```diff
@@ -264,6 +269,99 @@ class DynamicalCoreConfig:
...
+            f90_nml = f90nml.read(self.namelist_override)
+            namelist = Namelist.from_f90nml(f90_nml)
+            self.dt_atmos = namelist.dt_atmos
+            self.a_imp = namelist.a_imp
+            self.beta = namelist.beta
+            self.consv_te = namelist.consv_te
+            self.d2_bg = namelist.d2_bg
+            self.d2_bg_k1 = namelist.d2_bg_k1
+            self.d2_bg_k2 = namelist.d2_bg_k2
+            self.d4_bg = namelist.d4_bg
```
elynnwu, line 287: Is there a way to have post_init return a new class and override the initialization? I tried a few variations of calling class methods but no luck, so ended up settling on explicitly setting each variable.

driver/Makefile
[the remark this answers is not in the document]
elynnwu: Okay, `test_mpi_serialbox` removed from Makefile and moved the command directly to `test_driver.sh`. Keeping `baroclinic_c12_from_serialbox.yaml` in the examples/configs folder because it needs to be somewhere docker can access. I think it makes sense for the example yaml file to be there even though it points to a path that people not using docker will not have access, information on the path is added in README.

driver/examples/configs/README.md
```diff
@@ -0,0 +1,45 @@
+# Driver Configurations
```
mcgibbon, line 1: This is good to have for now, in the long term it will be hard to maintain this README separate from the code. If and when we put together documentation, we should try to make it so Sphinx automatically documents these details based on dataclass docstrings.

driver/examples/configs/baroclinic_c12_from_serialbox.yaml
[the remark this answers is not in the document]
mcgibbon: What you have right now (the config is here but not the make target) is good. You could also consider putting the config under `driver/tests/` or some other testing directory.

driver/pace/driver/run.py
```diff
@@ -116,85 +130,172 @@ class SerialboxConfig(InitializationConfig):
...
     def start_time(self) -> datetime:
         return datetime(2000, 1, 1)
 
+    @property
+    def f90_namelist(self) -> f90nml.Namelist:
+        return f90nml.read(self.path + "/input.nml")
+
+    @property
+    def namelist(self) -> Namelist:
+        return Namelist.from_f90nml(self.f90_namelist)
```
mcgibbon, line 145: Can these be private attributes?

driver/pace/driver/run.py
```diff
@@ -116,85 +130,172 @@ class SerialboxConfig(InitializationConfig):
...
+            )
+        return grid, grid_data, damping_coeff, driver_grid_data
+
     def get_driver_state(
         self,
         quantity_factory: pace.util.QuantityFactory,
         communicator: pace.util.CubedSphereCommunicator,
     ) -> DriverState:
-        raise NotImplementedError()
+        self.backend = quantity_factory.empty(
```
mcgibbon, line 250: Can you pass this as an explicit argument to the private methods that use it, instead of passing it through an object attribute? We don't need this data to persist on `self` between calls, and passing it as an argument makes this clearer.

driver/pace/driver/run.py
```diff
@@ -292,34 +394,50 @@ def from_dict(cls, kwargs: Dict[str, Any]) -> "DriverConfig":
...
+        kwargs["dycore_config"] = dacite.from_dict(
+            data_class=fv3core.DynamicalCoreConfig,
+            data=kwargs.get("dycore_config", {}),
+            config=dacite.Config(strict=True),
+        )
+        kwargs["physics_config"] = dacite.from_dict(
+            data_class=fv3gfs.physics.PhysicsConfig,
+            data=kwargs.get("physics_config", {}),
+            config=dacite.Config(strict=True),
+        )
```
mcgibbon, line 420: Is this necessary? We expect dycore_config and physics_config should always be included in the yaml.

driver/pace/driver/run.py
```diff
@@ -292,34 +394,50 @@ def from_dict(cls, kwargs: Dict[str, Any]) -> "DriverConfig":
...
+            kwargs["physics_config"].layout = kwargs["layout"]
+            kwargs["physics_config"].dt_atmos = kwargs["dt_atmos"]
+            kwargs["physics_config"].npx = kwargs["nx_tile"] + 1
+            kwargs["physics_config"].npy = kwargs["nx_tile"] + 1
+            kwargs["physics_config"].npz = kwargs["nz"]
+
+        elif initialization_type == "serialbox":
+            kwargs["nx_tile"] = kwargs["dycore_config"].npx - 1
+            kwargs["nz"] = kwargs["dycore_config"].npz
+            kwargs["layout"] = tuple(kwargs["dycore_config"].layout)
```
mcgibbon, line 439: This still makes me nervous (changing the behavior of the coupling when initialization_type="serialbox"), a consequence of that choice is this has pretty odd behaviors in the case where initialization_type="serialbox" and namelist_override is not used for one or both of dycore_config and physics_config. We don't plan to use it but it's a valid configuration option presented to the user.

Is it that bad to rely on the yaml for the top-level settings (nx_tile, nz, layout) and expect the user to make sure these correspond with the serialbox directory?

Admittedly this is a lot cleaner and less dangerous than before, so I'll leave it up to you. If you keep it this way, I would suggest adding a check that raises an exception if initialization_type="serialbox" and namelist_override is not given for either of dycore_config and physics_config.

fv3core/fv3core/_config.py
mcgibbon: I expected you'd set each variable here. If you like you can

1) make a new object in here with DynamicalCoreConfig.from_f90nml
2) iterate over the dataclass fields of the object or self (they're the same type) and assign with setattr and getattr, this is OK type-wise because you're assigning from and to the same type without adding new attributes.

fv3gfs-physics/fv3gfs/physics/_config.py
```diff
@@ -93,6 +98,77 @@ class PhysicsConfig:
     tice: float = NamelistDefaults.tice
     alin: float = NamelistDefaults.alin
     clin: float = NamelistDefaults.clin
+    namelist_override: str = None
+
+    def __post_init__(self):
+        if self.namelist_override is not None and os.path.exists(
+            self.namelist_override
```
mcgibbon, line 105: It would be safer to raise an exception if the override path does not exist.

mcgibbon (commented):
Great job! Could be mergeable with few or no changes, not approving yet in case you act on one or more of the change suggestions.

rheacangeo (approved):
lgtm

driver/pace/driver/run.py
elynnwu: Yes, forgot to update this now that we don't need to call them from outside the class.

driver/pace/driver/run.py
elynnwu: Done!

driver/pace/driver/run.py
elynnwu: It is necessary because we want to use nx_tile, nz, and layout from dycore_config.

driver/pace/driver/run.py
elynnwu: You can actually still have serialbox initialization and pass in dycore_config and physics_config. I think we don't expect people to do that (assuming if you have serialized data, you ran fortran with a nml file), but in theory, it is allowed and shouldn't break the code.

fv3core/fv3core/_config.py
elynnwu: Great suggestion, I knew there was a better way to utilize `from_f90nml`! Updating this.

driver/pace/driver/run.py
mcgibbon: What I mean is, the normal expectation here is that you don't pass `npx` in `dycore_config`, and instead you pass `nx_tile`. If you take a configuration file that does this and modify the initailization_type, npx will be set to 0 and no exception will be raised. This could be pretty hard to detect and debug.

mcgibbon: Overriding the owning class behavior based on the choice of subclass for initialization makes me nervous because of how easy it is to not notice potential bugs like that, and how hard it is to detect them when they do happen.

If we consider the "serialbox" initialization to be a use-at-your-own-risk configuration that may cause strange bugs then this is OK, but I'm not sure the tradeoff of not copying these variables from the namelist to the yaml outweighs the risk of us running into one of these bugs. We can merge this as-is if you think it will be fine.

mcgibbon: Actually, is there a more important bug here? Does `dt_atmos` as well as `days`, `hours`, `minutes`, `seconds` need to be copied from DycoreConfig to DriverConfig as well?

mcgibbon (approved):
LGTM, thanks for all the work avoiding side-effects on this one!

elynnwu:
launch jenkins

PR #109, opened 2022-02-10
Driver outputs performance timing

elynnwu:
## Purpose

We move the timing capability in standalone `dynamics.py` to driver. Driver now has the option to output the same json file. 

## Code changes:

- `driver/run.py`: added `PerformanceConfig` to switch on/off timing
- added `report.py` to `pace.util`: functions inside `dynamics.py` regarding timing information and output file are moved here. In addition, we also output SYPD. Format version is changed to 3 since it's not identical to the current `dynamics.py` output json. No git hash is passed to driver for now, when we decide to switch off dynamics in favor of driver, we can do it then.
- added `Barrier` in `NullComm` which behaves the same as `barrier` - it is used in driver to collect performance report

## Checklist


driver/pace/driver/run.py
```diff
@@ -459,47 +477,65 @@ def __init__(
...
-        quantity_factory, stencil_factory = _setup_factories(
-            config=config, communicator=communicator
-        )
+        self.comm = comm
+        if self.config.performance_config.timer:
+            self.timestep_timer = pace.util.Timer()
+            self.timer = pace.util.Timer()
+        else:
+            self.timestep_timer = pace.util.NullTimer()
+            self.timer = pace.util.NullTimer()
```
mcgibbon, line 486:
```suggestion
        if self.config.performance_config.timer:
            self.timestep_timer = pace.util.Timer()
            self.timer = pace.util.Timer()
        else:
            self.timestep_timer = pace.util.NullTimer()
            self.timer = pace.util.NullTimer()
```
You could move this concern out of driver and onto PerformanceConfig by making a property method `timer` on `PerformanceConfig` which gives a NullTimer in the right circumstances. Similarly for writing the results to disk.

driver/pace/driver/run.py
```diff
@@ -509,10 +545,39 @@ def step_all(self):
...
+            collect_data_and_write_to_file(
+                len(self.hits_per_step) - 1,
+                self.config.stencil_config.backend,
+                self.config.git_hash,
+                self.comm,
+                self.hits_per_step,
+                self.times_per_step,
+                self.config.experiment_name,
+                self.config.dt_atmos,
+            )
```
mcgibbon, line 580: Can these concerns be put on the performance class instead of the driver class?

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def gather_hit_counts(
+    hits_per_step: List[Dict[str, int]], results: Dict[str, Any]
+) -> Dict[str, Any]:
+    """collects the hit count across all timers called in a program execution"""
+    for data_point in hits_per_step:
+        for name, value in data_point.items():
+            if name not in results["times"]:
+                print(name)
```
mcgibbon, line 80: If there is a print statement you want to keep, can you use a logger instead?

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+            keys.add(k)
+    sorted_keys = list(keys)
+    sorted_keys.sort()
+    return sorted_keys
+
+
+def gather_timing_data(
+    times_per_step: List[Dict[str, float]],
+    results: Dict[str, Any],
+    comm: MPI.Comm,
```
mcgibbon, line 42: This type hint will cause an exception if mpi4py (an optional dependency) is not installed, in which case MPI will be None.

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+        if is_root:
+            results["times"][timer_name]["times"] = copy.deepcopy(recvbuf.tolist())
+    return results
+
+
+def write_global_timings(experiment: Dict[str, Any]) -> None:
+    now = datetime.now()
+    filename = now.strftime("%Y-%m-%d-%H-%M-%S")
+    with open(filename + ".json", "w") as outfile:
+        json.dump(experiment, outfile, sort_keys=True, indent=4)
```
mcgibbon, line 70: Other than the filename used, is there anything that makes this function specific to writing global timings? It seems like `write_to_timestamped_json` might be more descriptive.

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+import json
+from datetime import datetime
+from typing import Any, Dict, List
+
+import numpy as np
+
+from pace.util.mpi import MPI
+
+
+def set_experiment_info(
```
mcgibbon, line 11: The methods you added here don't have unit tests - can you add some?

Is there a reason to put these in `pace-util` or should they be put in `driver`? Some of these seem pretty specialized to timing the full model.

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+    experiment_name: str, time_step: int, backend: str, git_hash: str
+) -> Dict[str, Any]:
+    experiment: Dict[str, Any] = {}
+    now = datetime.now()
+    dt_string = now.strftime("%d/%m/%Y %H:%M:%S")
+    experiment["setup"] = {}
```
mcgibbon, line 17: There's a lot of use of nested dictionaries here - this might be easier to work with and harder to make bugs on if you turn this into a dataclass structure.

pace-util/pace/util/report.py
```diff
@@ -0,0 +1,114 @@
...
+        results = set_experiment_info(experiment_name, time_step, backend, git_hash)
+        results = gather_hit_counts(hits_per_step, results)
+
+    results = gather_timing_data(times_per_step, results, comm)
+
+    if is_root:
+        mainloop = np.mean(sum(results["times"]["mainloop"]["times"], []))
+        speedup = dt_atmos / mainloop
+        SYPD = 1.0 / 365.0 * speedup
+        results["SYPD"] = SYPD
```
mcgibbon, line 113: nit: Can you refactor this variable name to be lowercase? The syntax highlighting messes with me.

driver/pace/driver/run.py
```diff
@@ -353,6 +360,14 @@ class DriverConfig:
...
         layout: number of ranks along the x and y dimensions
         dt_atmos: atmospheric timestep in seconds
+        diagnostics_config: configuration for output diagnostics
+        experiment_name: name of this experiment
+        dycore_config: configuration for dynamical core
+        physics_config: configuration for physics
+        days: simulation time in days
+        hours: simulation time in hours
+        minutes: simulation time in minutes
+        seconds: simulation time in seconds
```
mcgibbon, line 370: Thanks for catching/fixing this!

driver/pace/driver/run.py
```diff
@@ -363,6 +378,8 @@ class DriverConfig:
     layout: Tuple[int, int]
     dt_atmos: float
     diagnostics_config: DiagnosticsConfig
+    performance_config: PerformanceConfig
+    experiment_name: str = "test"
```
mcgibbon, line 382: Should experiment_name and git_hash be part of PerformanceConfig? They aren't used otherwise in Driver.

Is it possible to programmatically determine the git hash inside Python, instead of passing it as a yaml/command-line configuration value?

mcgibbon (changes requested):
There are some tests needed for parts of this code, which we'll need to be confident in its results, and there's one bug involving MPI=None. Apart from those I put (possibly too many) hopefully helpful suggestions.

driver/pace/driver/run.py
elynnwu: Great idea! This is now moved to `PerformanceConfig`.

driver/pace/driver/run.py
elynnwu: Moved to `PerformanceConfig`.

pace-util/pace/util/report.py
elynnwu: I think this is leftover from before, removing print here.

pace-util/pace/util/report.py
elynnwu: Removed type hint here.

pace-util/pace/util/report.py
elynnwu: Changed to `write_to_timestamped_json`

driver/pace/driver/run.py
elynnwu: Moved `experiment_name` to `PerformanceConfig`. `git_hash` is determined from within python;

pace-util/pace/util/report.py
elynnwu: Moved report.py to driver

pace-util/pace/util/report.py
elynnwu: Done!

pace-util/pace/util/report.py
elynnwu: Added unit test: test_timing_info

driver/examples/configs/baroclinic_c12_from_serialbox.yaml
```diff
@@ -8,6 +8,9 @@ initialization_type: serialbox
 initialization_config:
   path: /test_data/c12_6ranks_baroclinic_dycore_microphysics
   serialized_grid: True
+performance_config:
+  timer: true
```
twicki, line 12: is this not always true for any performance run?  would it help if we just had a way to run in performance mode and have the performance config default to true/true?

driver/examples/configs/baroclinic_c12.yaml
```diff
@@ -6,6 +6,10 @@ stencil_config:
   device_sync: true
 initialization_type: baroclinic
 initialization_config: {}
+performance_config:
+  timer: true
+  json_dump: true
+  experiment_name: c12_baroclinic
```
twicki, line 12: why is the experiment name specific to the performance config?

driver/pace/driver/run.py
```diff
@@ -1,6 +1,8 @@
 import abc
 import dataclasses
 import functools
+import os
+import subprocess
```
twicki, line 5: I generally try to avoid subprocess for something as simple as getting a git hash (just because it is somewhat finicky). Why did you opt for it to fetch it from here compared to as an input arg?

driver/pace/driver/run.py
```diff
@@ -27,6 +29,8 @@
 from pace.util.grid import DampingCoefficients
 from pace.util.namelist import Namelist
 
+from .report import collect_data_and_write_to_file
```
twicki, line 32: why are you opting for relative imports here?

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@dataclasses.dataclass
+class Experiment:
+    dataset: str
+    format_version: int
+    git_hash: str
+    timestamp: str
+    timesteps: int
+    version: str
```
twicki, line 19: Version is a strange naming choice here (especially with format version present). Why is this not called backend?

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+class TimeReport:
+    hits: int
+    times: list
+
+
+@dataclasses.dataclass
+class Report:
+    setup: Experiment
+    times: dict
+    SYPD: float
```
twicki, line 32: would it not be more interesting to store `dt_atmos` and `__post_init__` calculate the SYPD?

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+    sorted_keys.sort()
+    return sorted_keys
+
+
+def gather_timing_data(
+    times_per_step: List[Dict[str, float]],
+    comm,
+    root: int = 0,
+) -> Dict[str, Any]:
+    """returns an updated version of  the results dictionary owned
```
twicki, line 65:
```suggestion
    """returns an updated version of the results dictionary owned
```

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+    """collects the hit count across all timers called in a program execution"""
+    for data_point in hits_per_step:
+        for name, value in data_point.items():
+            timing_info[name].hits += value
+    return timing_info
+
+
+def get_sypd(timing_info: Dict[str, TimeReport], dt_atmos: float) -> float:
+    if "mainloop" in timing_info:
+        mainloop = np.mean(sum(timing_info["mainloop"].times, []))
```
twicki, line 108: why is this the sum of all the mainloop times?

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def make_report(
+    exp_info: Experiment,
+    timing_info: Dict[str, TimeReport],
+    dt_atmos: float,
+):
+    sypd = get_sypd(timing_info, dt_atmos)
+    report = Report(setup=exp_info, times=timing_info, SYPD=sypd)
+    return dataclasses.asdict(report)
```
twicki, line 123: if we already have the dataclass, why do we make it a dict here? why not let that be the responsibility of the function consuming the a dict?

driver/pace/driver/report.py
```diff
@@ -0,0 +1,146 @@
...
+    hits_per_step: List,
+    times_per_step: List,
+    experiment_name: str,
+    dt_atmos: float,
+) -> None:
+    """
+    collect the gathered data from all the ranks onto rank 0 and write the timing file
+    """
+    is_root = comm.Get_rank() == 0
+    timing_info = gather_timing_data(times_per_step, comm)
```
twicki, line 140: is there never need for a non-MPI version of this?

driver/pace/driver/run.py
```diff
@@ -459,47 +549,55 @@ def __init__(
...
-            comm=comm, layout=self.config.layout
-        )
-        quantity_factory, stencil_factory = _setup_factories(
-            config=config, communicator=communicator
-        )
+        self.comm = comm
+        self.performance_config = self.config.performance_config
+        self.timestep_timer = self.performance_config.timestep_timer
+        self.timer = self.performance_config.total_timer
+        self.times_per_step = self.performance_config.times_per_step
```
twicki, line 556: why are we propagating these variables out here?

driver/pace/driver/run.py
```diff
@@ -459,47 +549,55 @@ def __init__(
...
+        self.hits_per_step = self.performance_config.hits_per_step
+        self.timer.start("total")
```
twicki, line 558: how dangerous is it that this timer starts in one method and is stopped in another one?

driver/pace/driver/run.py
mcgibbon: Because I asked for it. What's finicky about subprocess?

driver/pace/driver/run.py
mcgibbon: Might be other reasons, but also to some degree because I asked for it. I think either should be fine for anyone to merge, but relative is nice in that it clues you in to "this is from the same package", especially when we have a multi-package structure for our code.

driver/examples/configs/baroclinic_c12_from_serialbox.yaml
elynnwu: This is a good point. Changing it to just one config.

driver/examples/configs/baroclinic_c12.yaml
elynnwu: This was a suggestion by Jeremy because experiment name is only used inside performance config. We do not use `input.yml`, the only time that comes into play is if we want to initialize with serialbox (which is expected to only be used internally).

driver/pace/driver/report.py
elynnwu: oh right, this sum is because `timing_info` has nested list like this: `[[timestep1_rank0, timestep2_rank0, timestep3_rank0], [timestep1_rank1, timestep2_rank1, timestep3_rank1], ...]`. So the sum over `[]` collapses everything into one long list.

driver/pace/driver/report.py
elynnwu: Good point, moving this to when the data is written, and remove `make_report`

driver/pace/driver/report.py
elynnwu: I don't think so, because driver takes in mpi4py MPI.

elynnwu: Do we want to support non-MPI version of driver? @mcgibbon

driver/pace/driver/run.py
elynnwu: Moved them to class attributes instead of properties.

driver/pace/driver/run.py
elynnwu: Yes either works, will keep it as relative import unless there's a strong preference for absolute.

driver/pace/driver/report.py
elynnwu: Changed to backend, I was trying to not change anything so we didn't have to update the viz plan. But looks like we would need to anyways because I can't use `hash` and had to change it to `git_hash`.

driver/pace/driver/report.py
elynnwu: That makes sense, changed to calculate SYPD in post init.

driver/pace/driver/run.py
elynnwu: Could be dangerous, but we still have a null timer when performance mode is false. 

The other option is to not have total time at all, I think initialization is what we're after. The total time should be the same as initialization + all mainloop + time to write out performance output file (which we do not time), but when submitted through slurm, we should have the real total time anyways. I was trying to keep ~identical to `dynamics.py`, but we can definitely change that.

driver/pace/driver/report.py
mcgibbon: Can you write a short unit test for this? I also don't understand how this code will do what I'd expect (what even is `sum(list_of_lists, [])`? sum() returns a float, so why the `np.mean`?).

driver/pace/driver/report.py
mcgibbon: The code you have as-written will support non-MPI. When we run in non-MPI mode, we still have a `comm` object, it's just a `LocalComm` instead of an mpi4py comm object. Get_rank(), gather, scatter, etc. all should work just fine without MPI so long as we use the passed `comm` object instead of accessing mpi4py directly.

driver/pace/driver/run.py
```diff
@@ -508,10 +587,31 @@ def step_all(self):
...
+            self.performance_config.hits_per_step,
+        )
+        self.performance_config.write_out_performance(
+            self.comm,
+            self.config.stencil_config.backend,
+            self.config.dt_atmos,
+            self.performance_config.total_timer,
+            self.performance_config.times_per_step,
+            self.performance_config.hits_per_step,
+        )
```
mcgibbon, line 603: Just realized - I wouldn't expect `step_all` to write out performance. Can this be moved to another method, invoked from the top level?

Very optional: If you do this, you could also invoke the timer.start("total") and timer.stop("total") (or use `with`) from the top level `main` function, which would be safer than invoking it in two different methods. I think deleting the "total" times is also an option, I don't see the use case if we already have an initialization + call timing.

Is it an issue that we're timing the first timestep in the same place as the rest? I think we do some compilation at that time.

driver/pace/driver/report.py
elynnwu: `sum(list_of_lists, [])` returns a flatten list. Added a unit test: `test_sypd`

driver/pace/driver/run.py
elynnwu: Moved write out performance to be invoked from `main`. Total time is now moved to `main`, but it is the time it takes for `step_all`. I'm also good with just deleting this completely, let's see what @twicki says.

driver/pace/driver/run.py
```diff
@@ -508,10 +586,22 @@ def step_all(self):
             self._step(timestep=self.config.timestep.total_seconds())
             time += self.config.timestep
             self.diagnostics.store(time=time, state=self.state)
+        self.performance_config.collect_performance(
+            self.performance_config.timestep_timer,
+            self.performance_config.times_per_step,
+            self.performance_config.hits_per_step,
+        )
```
mcgibbon, line 593: If you'd like to avoid passing these as arguments twice, you could use the attributes on `self` in a `performance_config.collect_performance()`.

driver/pace/driver/run.py
```diff
@@ -508,10 +586,22 @@ def step_all(self):
...
-        self._step_physics(timestep=timestep)
+        with self.performance_config.timestep_timer.clock("mainloop"):
+            self._step_dynamics(timestep=timestep)
+            self._step_physics(timestep=timestep)
+        self.performance_config.collect_performance(
+            self.performance_config.timestep_timer,
+            self.performance_config.times_per_step,
+            self.performance_config.hits_per_step,
+        )
+        self.performance_config.timestep_timer.reset()
```
mcgibbon, line 604: Should `reset` be called automatically inside `collect_performance`? I see you call it here but not in `step_all`.

driver/pace/driver/run.py
```diff
@@ -533,6 +624,16 @@ def _step_physics(self, timestep: float):
...
         )
 
+    def write_performance_json_output(self):
+        self.performance_config.write_out_performance(
+            self.comm,
+            self.config.stencil_config.backend,
+            self.config.dt_atmos,
+            self.performance_config.total_timer,
+            self.performance_config.times_per_step,
+            self.performance_config.hits_per_step,
```
mcgibbon, line 634: Similarly you could use `self` attributes for things existing on `performance_config` here. It's reasonable that if you write performance from the object it should write the performance it's tracking.

driver/pace/driver/run.py
elynnwu: The last one wasn't reset since it's not used again, but it is not a problem if it gets reset either. Moved `reset` to `collect_performance`

driver/pace/driver/run.py
elynnwu: Done for both. 
I originally had `*_per_step` as properties, so had to keep track of them from within driver class. After Tobias' suggestion, I moved them to be performance config class attributes, and things are definitely way cleaner.

driver/pace/driver/run.py
```diff
@@ -311,6 +315,62 @@ class DiagnosticsConfig:
...
+    total_timer: pace.util.Timer = pace.util.NullTimer()
+    times_per_step: List = dataclasses.field(default_factory=list)
+    hits_per_step: List = dataclasses.field(default_factory=list)
+
+    def __post_init__(self):
+        if self.performance_mode:
+            self.timestep_timer = pace.util.Timer()
+            self.total_timer = pace.util.Timer()
+
+    def collect_performance(self):
```
mcgibbon, line 332:
```suggestion
    def collect_performance(self):
        """
        Take the accumulated timings and flush them into a new entry
        in times_per_step and hits_per_step.
        """
```

driver/pace/driver/run.py
```diff
@@ -577,7 +659,9 @@ def main(driver_config: DriverConfig, comm):
         config=driver_config,
         comm=comm,
     )
-    driver.step_all()
+    with driver.performance_config.total_timer.clock("total"):
+        driver.step_all()
```
mcgibbon, line 663: nit: if this is purely timing `step_all` I think you can move it inside that routine. This would avoid having a doubly-nested access (driver.<object>.<sub-object>.<method>). It only needed to be at a higher level earlier when you were timing across methods.

pace-util/HISTORY.md
```diff
@@ -17,6 +17,8 @@ Minor changes:
 - added `CubedSphereCommunicator.from_layout` constructor method
 - added support for built-in `datetime` in ZarrMonitor
 - `edge_interior_ratio` is now an optional argument of `tile_extent_from_rank_metadata`
+- added `report` for saving performance timing outputs
+- added `Barrier` in `NullComm` which behaves the same as `barrier`
```
mcgibbon, line 21: NullComm is in `fv3core` and report is in `driver`, so you don't need these in the pace-util history.

We could move NullComm to `pace-util` later but report definitely seems to be long in `driver`.

```suggestion
```

mcgibbon (approved):
Thanks! Some small doc fixes/additions suggested. Only required change is to remove the non-pace-util changes from its HISTORY file.

driver/pace/driver/run.py
```diff
@@ -311,6 +315,62 @@ class DiagnosticsConfig:
...
+                        ["git", "-C", driver_path, "rev-parse", "HEAD"]
+                    )
+                    .decode()
+                    .rstrip()
+                )
+            except subprocess.CalledProcessError:
+                git_hash = "notarepo"
+
+            self.times_per_step.append(self.total_timer.times)
+            self.hits_per_step.append(self.total_timer.hits)
```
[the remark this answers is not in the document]
elynnwu: The problem is that `timestep_timer` gets reset every step, but `total_timer` cannot. This is here to tag on the total timer results without having to keep two lists and make sure that the format is the same.

PR #114, opened 2022-02-14
fix daint compilation

rheacangeo:
## Purpose
Reordering package imports in stencil.py so that pace.util loads after gt4py.<features>, otherwise on daint compilation breaks with 
`Field decl type <class 'gtc.gtir.FieldDecl'> has no 'dimensions' attribute `
from the fv3core standalone. 
Another solution that works is removing the imports from dsl/pace/dsl/__init__.py, but having those makes the driver cleaner to read.

### DIFF
--- a/dsl/pace/dsl/stencil.py
+++ b/dsl/pace/dsl/stencil.py
@@ -22,9 +22,9 @@ import numpy as np
 from gt4py import gtscript
 from gt4py.storage.storage import Storage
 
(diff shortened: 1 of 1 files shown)


mcgibbon (approved):
Good job figuring this out, I can’t imagine it was easy.

rheacangeo:
launch jenkins

PR #117, opened 2022-02-16
another stab at fixing the xarray/gt4py import issue

rheacangeo:
## Purpose
The last change (#114) had worked locally on daint, but broke when run on jenkins performance plans. Here is another attempt to resolve the xarray/gt4py import order that also seems to 'fix' the issue locally. Before merging, will test this branch on jenkins.


rheacangeo:
Update, this also repeatedly works as my user on daint, but not for jenkins.

PR #125, opened 2022-02-23
Driver diagnostics write out lat/lon info

elynnwu:
## Purpose

Driver diagnostics now writes out latitude and longitude, and we can plot cube with driver output.

## Code changes:

- `postprocessing.Dockerfile`: switch to ubuntu 22, this is now supported on daint (through sarus)
- `plot_cube.py`: an example script to plot driver output using `fv3viz`
- `driver/run.py`: add `write_grid` option to `store`, which is currently hard coded to write out only `lat` and `lon` (at the interface as part of the requirement in `fv3viz`)
- `zarr_monitor.py`: add `_ZarrGridWriter` to support writing grid data (write only once, without time coordinate)

## Requirements changes:

- [optional] `fv3viz` is used in `plot_cube.py`: this is not a requirement in pace but users have the option to follow `fv3net` requirements or use the post processing docker file

Example c48 driver output:

## Infrastructure changes:

- Post processing dockerfile now uses ubuntu 22

## Checklist

### DIFF
--- a/docker/postprocessing.Dockerfile
+++ b/docker/postprocessing.Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:22.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
@@ -11,49 +11,20 @@ RUN apt-get update -y && \
     g++ \
     gcc \
     gfortran \
-    make \
-    wget \
-    pkg-config \
-    libsqlite3-dev \
-    sqlite3 \
-    libcurl4-openssl-dev \
-    libtiff5 \
-    libtiff5-dev \
-    openssl \
-    libssl-dev
-
-RUN cd / && \
-    wget https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2.tar.gz && \
-    tar xzf cmake-3.22.2.tar.gz && \
-    rm cmake-3.22.2.tar.gz && \
-    cd cmake-3.22.2 && \
-    ./bootstrap && make -j4 && make install
-
-RUN cd / && \
-    wget http://download.osgeo.org/geos/geos-3.8.2.tar.bz2 && \
-    tar xfj geos-3.8.2.tar.bz2 && \
-    rm geos-3.8.2.tar.bz2 && \
-    cd geos-3.8.2 && \
-    mkdir _build && \
-    cd _build && \
-    cmake .. && make -j4 && ctest && make install
-
-RUN cd / && \
-    wget http://archive.ubuntu.com/ubuntu/pool/universe/p/proj/proj_8.2.1.orig.tar.gz && \
-    tar xzf proj_8.2.1.orig.tar.gz && \
-    rm proj_8.2.1.orig.tar.gz && \
-    cd proj-8.2.1 && \
-    ./configure && make -j4 && make install
+    libproj-dev \
+    proj-data \
+    proj-bin \
+    libgeos-dev
 
 RUN apt-get update -y && \
     apt install -y --no-install-recommends \
     git \
-    python3.8 \
-    python3.8-dev &&\
+    python3.9 \
+    python3.9-dev &&\
     rm -rf /var/lib/apt/lists/* && \
     apt-get clean
 
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 60
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.9 60
 
 RUN apt-get update -y &&\
     apt install -y --no-install-recommends\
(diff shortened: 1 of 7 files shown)


driver/examples/plot_cube.py
```diff
@@ -0,0 +1,22 @@
+import matplotlib.pyplot as plt
```
mcgibbon, line 1: later: If we're going to be plotting cube data a lot, we might consider copy-pasting `pcolormesh_cube` into one of the packages in `pace` so it's easier for users to set up their environment.

docker/postprocessing.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:22.04
```
mcgibbon, line 1: If you're spending a lot of time managing this dockerfile, I'd suggest you use one of the fv3net-published ones instead. As it is this is duplicated work from what the ML team is already doing. These are pushed to dockerhub on every PR and tagged by fv3net git sha.

driver/pace/driver/run.py
```diff
@@ -324,7 +324,7 @@ def __init__(
             store=store, partitioner=partitioner, mpi_comm=comm
         )
 
-    def store(self, time: datetime, state: DriverState):
+    def store(self, time: datetime, state: DriverState, write_grid=False):
```
mcgibbon, line 327: `store`'s responsibilities become a lot more complicated with this new option, since state wants to be stored every step and grid wants to be stored only once. Can you move grid writing into its own method?

pace-util/pace/util/zarr_monitor.py
```diff
@@ -102,6 +103,7 @@ def _check_writers(self, state):
                 "that were not present in earlier states"
             )
         missing_names = set(self._writers.keys()).difference(state.keys())
+        missing_names.difference_update(self._bypass_checks)
```
mcgibbon, line 106: Grid data only needs to be written once, so you shouldn't need to keep them in `self._writers`. If you never append them into `self._writers` you don't need the _bypass_checks logic.

pace-util/pace/util/zarr_monitor.py
```diff
@@ -126,6 +128,18 @@ def store(self, state: dict) -> None:
         for name, quantity in sorted(state.items(), key=lambda x: x[0]):
             self._writers[name].append(quantity)  # type: ignore[index]
 
+    def store_grid(self, grid: dict) -> None:
```
mcgibbon, line 131: Can you remove the name reference to "grid" in this file, and instead use a generic term like "constant"? This logic should work for anything that isn't time-varying.

driver/pace/driver/run.py
elynnwu: Changed to `store_grid`

pace-util/pace/util/zarr_monitor.py
elynnwu: Removed from `self._writers`

pace-util/pace/util/zarr_monitor.py
elynnwu: Changed to `store_constant`

docker/postprocessing.Dockerfile
elynnwu: That is a good point, did not realize fv3net has that. I was just trying to keep it as minimal as possible because this needs to be pulled from daint through sarus, and I was running into some problems with libraries when I had more requirements. I think for now we can just use this, but if it becomes a problem we should look into switching to fv3net instead of having to maintain this.

pace-util/pace/util/zarr_monitor.py
```diff
@@ -126,6 +126,16 @@ def store(self, state: dict) -> None:
         for name, quantity in sorted(state.items(), key=lambda x: x[0]):
             self._writers[name].append(quantity)  # type: ignore[index]
 
+    def store_constant(self, grid: dict) -> None:
+        for name, quantity in grid.items():
```
mcgibbon, line 130: Might need some imports to make this code suggestion work. You also might want to test what happens when you do this, I'm not sure if it will error (due to the store already being created) or overwrite it. If it errors you could consider adding some kind of running tab of what constants have been stored, and raising a more helpful exception instead of adding this docstring note. My note isn't great because if there's a docstring about the errors they aren't exactly "unexpected".

```suggestion
    """
    Store a state dictionary containing values which do not vary in time.
    
    This function can be called at most once for a given state key,
    if you try to store the same constant multiple times you may
    get unexpected errors.
    """
    def store_constant(self, state: Mapping[str, Quantity]) -> None:
        for name, quantity in state.items():
```

pace-util/pace/util/zarr_monitor.py
```diff
@@ -259,6 +269,51 @@ def _get_from_slice(target_slice):
...
+                self.array[target_slice] = cupy.asnumpy(quantity.view[:][from_slice])
+            else:
+                raise err
+        except TypeError as err:
+            if err.args[0].startswith(
+                "Implicit conversion to a NumPy array is not allowed."
+            ):
+                self.array[target_slice] = cupy.asnumpy(quantity.view[:][from_slice])
+            else:
+                raise err
```
mcgibbon, line 314: Would have been ideal to not have to copy all this code, but this is fine.

mcgibbon (approved):
Thanks for adding this feature! `store_constant` could use a docstring, otherwise looks ready to merge!

elynnwu:
launch jenkins

elynnwu:
launch jenkins

PR #128, opened 2022-02-24
Daint compilation

rheacangeo:
## Purpose
To re-enable compilation on daint after its environment changed. 

## Code changes:

- add an import of dsl.stencil to dynamics.py as the first thing to prevent the Field has no decl error
- Added back the single node compilation step in run_on_daint because nvcc is now crashing on multinode again (using gcc 9.3, cuda 11.2)


elynnwu (approved):
lgtm! Thanks for getting CI back!

PR #129, opened 2022-02-28
merge main_uc and main_vc stencils in fxadv

mcgibbon:
## Purpose

This PR merges two stencils in fxadv.py to improve performance.

## Code changes:

- Merged main_uc_stencil and main_vc_stencil in fxadv.py
- Remove two data copy operations
- No longer call edge/corner stencils if we aren't on an edge/corner rank in fxadv.py
- Disable validation of FxAdv on outermost halo for uc_contra and vc_contra

## Checklist

### DIFF
--- a/fv3core/fv3core/stencils/fxadv.py
+++ b/fv3core/fv3core/stencils/fxadv.py
@@ -8,12 +8,15 @@ from pace.util.grid import GridData
 
 # TODO: the mix of local and global regions is strange here
 # it's a workaround to specify DON'T do this calculation if on the tile edge
-def main_uc_contra(
+def main_uc_vc_contra(
     uc: FloatField,
     vc: FloatField,
     cosa_u: FloatFieldIJ,
     rsin_u: FloatFieldIJ,
+    cosa_v: FloatFieldIJ,
+    rsin_v: FloatFieldIJ,
     uc_contra: FloatField,
+    vc_contra: FloatField,
 ):
     """
     Args:
@@ -21,9 +24,12 @@ def main_uc_contra(
         vc (in): covariant c-grid y-wind
         cosa_u (in): ???
         rsin_u (in): ???
+        cosa_v (in): ???
+        rsin_v (in): ???
         uc_contra (out): contravariant c-grid x-wind
+        vc_contra (out): contravariant c-grid y-wind
     """
-    from __externals__ import j_end, j_start, local_ie, local_is
+    from __externals__ import j_end, j_start, local_ie, local_is, local_je, local_js
 
     with computation(PARALLEL), interval(...):
         utmp = uc_contra
@@ -35,6 +41,14 @@ def main_uc_contra(
             region[:, j_start - 1 : j_start + 1], region[:, j_end : j_end + 2]
         ):
             uc_contra = utmp
+
+        vtmp = vc_contra
+        with horizontal(region[:, local_js - 1 : local_je + 3]):
+            # for C-grid, u must be regridded to lie at same point as v
+            u = 0.25 * (uc[0, -1, 0] + uc[1, -1, 0] + uc + uc[1, 0, 0])
+            vc_contra = contravariant(vc, u, cosa_v, rsin_v)
+        with horizontal(region[:, j_start], region[:, j_end + 1]):
+            vc_contra = vtmp
 
 
 def uc_contra_y_edge(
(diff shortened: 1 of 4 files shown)


fv3core/tests/savepoint/test_translate.py
```diff
@@ -116,7 +116,7 @@ def sample_wherefail(
                 f"absolute diff {abs_err:.3e}, "
                 f"metric diff: {metric_err:.3e}"
             )
-        if metric_err > worst_metric_err:
+        if np.isnan(metric_err) or (metric_err > worst_metric_err):
```
mcgibbon, line 119: The test script was not giving output if all errors were nans, this fixes it.

fv3core/fv3core/stencils/fxadv.py
```diff
@@ -655,6 +642,8 @@ def __call__(
             vc_contra,
             dt,
         )
+        # self._set_nans(uc_contra)
+        # self._set_nans(vc_contra)
```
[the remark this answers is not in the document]
mcgibbon: Because this set_nans call is only useful in translate tests that are larger in scope than fxadv, like FVDynamics. 

The fxadv translate class will strip these nans off when testing. If I put this in TranslateFxAdv, the code would look like:
```
uc_contra = set_nans(uc_contra)  # sets nans in outermost halo
uc_contra = subselect(uc_contra)  # excludes outermost halo
```
in that case, setting nans carries no meaning because they're immediately discarded.

We want to see if these nans propagate into compute-domain information in one of the higher-level classes that rely on fxadv, like TranslateFVDynamics. In TranslateFVDynamics, we never see uc_contra from inside the translate class, so there's no opportunity to set nans in the center of the execution apart from dynamically modifying the class behavior (which dace doesn't like) or having code inside the compute class (which I did here).

fv3core/fv3core/stencils/fxadv.py
```diff
@@ -590,54 +580,51 @@ def __call__(
...
-        self._vc_contra_corners_stencil(
-            self._cosa_u,
-            self._cosa_v,
-            uc,
-            vc,
-            uc_contra,
-            vc_contra,
-            vc_contra,
-        )
+        if not self._tile_interior:
```
elynnwu, line 593: May not be important here, but I was wondering if having the if statement in python vs. in stencil makes any difference?

fv3core/fv3core/utils/functional_validation.py
```diff
@@ -0,0 +1,39 @@
...
+    dims: Sequence[str],
+    n_halo: Tuple[Tuple[int, int], Tuple[int, int]] = ((0, 0), (0, 0)),
+) -> Callable[[np.ndarray], np.ndarray]:
+    subset = get_subset_func(grid_indexing=grid_indexing, dims=dims, n_halo=n_halo)
+
+    def set_nans(data):
+        safe = copy.deepcopy(data)
+        data[:] = np.nan
+        data_subset = subset(data)
+        data_subset[:] = subset(safe)
```
elynnwu, line 37: I don't fully understand this function. Why do you need `data_subset` here if it's neither an argument or return?

fv3core/fv3core/stencils/fxadv.py
```diff
@@ -36,6 +42,11 @@ def main_uc_contra(
         ):
             uc_contra = utmp
 
+        with horizontal(region[:, local_js - 1 : local_je + 3]):
+            # for C-grid, u must be regridded to lie at same point as v
+            u = 0.25 * (uc[0, -1, 0] + uc[1, -1, 0] + uc + uc[1, 0, 0])
+            vc_contra = contravariant(vc, u, cosa_v, rsin_v)
```
elynnwu, line 48: How come you don't need to keep this for v but you need it for u above:

```        
with horizontal(region[:, j_start], region[:, j_end + 1]):
    vc_contra = vtmp
```

elynnwu (commented):
Nice to see that we can merge these! I have some clarification questions.

fv3core/fv3core/stencils/fxadv.py
mcgibbon: It should help here, we avoid a lot of python calls on top of stencil launch overheads by turning six stencil calls into one if statement.

I'd only put the if statement here because we already have separate stencils for edges and corners. If we had to separate stencils in order to use python if statements, it's much better to have joined stencils with the if statements inside.

fv3core/fv3core/utils/functional_validation.py
mcgibbon: Data_subset is a view into `data`. By modifying `data_subset`, we modify `data`. I'll add this as a comment.
```suggestion
    def set_nans(data):
        safe = copy.deepcopy(data)
        data[:] = np.nan
        # data_subset is a view of data, so modifying data_subset modifies data
        data_subset = subset(data)
        data_subset[:] = subset(safe)
```

fv3core/fv3core/stencils/fxadv.py
mcgibbon: That's a good question I only have a partial answer for. Removing this for this block only changes values in the outermost halo, while removing this for `u` also modifies eight points in the second ring of halos. Why this is, I can't say. Because we depend computationally on other points in the second ring of halos, it's more complicated to enable selective validation for that case.

elynnwu (approved):

PR #132, opened 2022-03-01
fix mirror_grid for even layouts

rheacangeo:
## Purpose
The 'mirror' code was rewritten from the tile-global loop to be computed per rank. Because we only tested with odd-numbered layouts, we had an error for the case where there are an even number of ranks in a tile, meaning the dateline/greenwich-meridian consistency correction needs to happen at the side of rank rather than in the middle. This correction has been tested on 6, 24, 54, and 96 ranks. 96 ranks has a 1e-11 threshold error, but this is unlikely related to the mirroring. 
## Code changes:

- mirror_grid corrections
- generation of grid update to call to mirror_grid


oelbert (approved):
Thanks for jumping on this!

rheacangeo:
launch jenkins

rheacangeo:
launch jenkins

PR #133, opened 2022-03-01
single node compilation

rheacangeo:
## Purpose

Since the daint upgrade, we haven't been able to compile multi-node, which could either be due to an issue with our distributed compilation, or a bug in the new cuda stack (gcc 10 no longer works with cuda 11.2). 
This introduces a dycore compiling script compile.py that will take the namelist we are testing and compile all the ranks in the first tile sequentially. For 6 ranks we can overload a node and do single node compilation, but this turned out to be slower than running them sequentially. But if we want to avoid running compile.py in our CI, I can branch on the number of ranks. 

## Code changes:

- added compile.py that runs the dycore with a null communicator and remapping replaced (to avoid the while loop running off the top of data that is nans) 
- updated run_on_daint to first run compile.py, then run the performance test.

### DIFF
--- a/fv3core/examples/standalone/runfile/compile.py
+++ b/fv3core/examples/standalone/runfile/compile.py
@@ -73,4 +73,4 @@ if __name__ == "__main__":
         )
         dycore, dycore_args = setup_dycore(dycore_config, mpi_comm, args.backend)
         with no_lagrangian_contributions(dynamical_core=dycore):
-            dycore.step_dynamics(*dycore_args)
+            dycore.step_dynamics(**dycore_args)
--- a/fv3core/examples/standalone/runfile/compile.py
+++ b/fv3core/examples/standalone/runfile/compile.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-import contextlib
 import os
 import sys
 from argparse import ArgumentParser, Namespace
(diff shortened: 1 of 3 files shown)


rheacangeo:
I verified this works for 54 ranks as well (except the compile job needs more than 2 hours, more like 4.5). Then the run.daint.slurm script takes 10 minutes.

fv3core/examples/standalone/runfile/compile.py
```diff
@@ -0,0 +1,86 @@
...
+    for rank in range(dycore_config.layout[0] * dycore_config.layout[1]):
+        mpi_comm = NullComm(
+            rank=rank,
+            total_ranks=6 * dycore_config.layout[0] * dycore_config.layout[1],
+            fill_value=0.0,
+        )
+        dycore, dycore_args = setup_dycore(
+            dycore_config, mpi_comm, args.backend, is_baroclinic_test_case
+        )
+        with no_lagrangian_contributions(dynamical_core=dycore):
```
twicki, line 84: because compilation happens at init-time I don't think we ever need to call the dycore

fv3core/examples/standalone/runfile/dynamics.py
```diff
@@ -214,6 +206,67 @@ def collect_data_and_write_to_file(
         write_global_timings(results)
 
 
+def setup_dycore(
```
twicki, line 209: this cleanup was very overdue, thanks

twicki (changes requested):

fv3core/examples/standalone/runfile/compile.py
rheacangeo: yay!

twicki (approved):
lgtm

PR #142, opened 2022-03-11
Feature/DaCe to pace savepoint test infrastructure

elynnwu:
## Purpose

Describe the purpose of this PR and the motivation if relevant to understanding. Include links to related issues, bugs or features.

Remove the sections below which do not apply.

## Code changes:

- Provide a list of relevant code changes and the side effects they have on other code.

## Requirements changes:

- Provide a list of any changes made to requirements, e.g. changes to files requirements*.txt, constraints.txt, setup.py, pyproject.toml, pre-commit-config.yaml and a reason if not included in the Purpose section (e.g. incompatibility, updates, etc)

## Infrastructure changes:

- Provide a list of changes that impact the infrastructure around running the code -- that is, changes to Makefiles, docker files, git submodules, or .jenkins (testing infrastructure changes). If Jenkins plans are also being manually changed, indicate that as well.

## Checklist

### DIFF
--- a/fv3core/.jenkins/actions/run_regression_tests.sh
+++ b/fv3core/.jenkins/actions/run_regression_tests.sh
@@ -2,8 +2,17 @@
 set -e -x
 BACKEND=$1
 EXPNAME=$2
+SAVEPOINT=${3:-All}
+ORCHESTRATION=${4:-none}
 XML_REPORT="sequential_test_results.xml"
-export TEST_ARGS="-v -s -rsx --backend=${BACKEND} "
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+JENKINS_DIR="$(dirname "$SCRIPT_DIR")"
+
+if [ $SAVEPOINT == "All" ]; then
+    export TEST_ARGS="-v -s -rsx --backend=${BACKEND} "
+else
+    export TEST_ARGS="-v -s -rsx --backend=${BACKEND} --which_modules=${SAVEPOINT}"
+fi
 
 JENKINS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )/../"
 
@@ -15,5 +24,9 @@ if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="srun" make tests savepoint_tests
 else
     export TEST_ARGS="${TEST_ARGS} --junitxml=/.jenkins/${XML_REPORT}"
-    VOLUMES="-v ${pwd}/.jenkins:/.jenkins" make tests savepoint_tests
+    if [ ${ORCHESTRATION} != "none" ]; then
+        VOLUMES="-v ${SCRIPT_DIR}/../:/.jenkins" DACE_INSTALL="/.jenkins/install_dace.sh ${ORCHESTRATION} &&" make savepoint_tests
+    else
+        VOLUMES="-v ${SCRIPT_DIR}/../:/.jenkins" make tests savepoint_tests
+    fi
 fi
--- a/fv3core/.jenkins/install_dace.sh
+++ b/fv3core/.jenkins/install_dace.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -e -x
+orchestration=$1
+if [ -d "gt4py" ]; then rm -rf gt4py; fi
+git clone https://github.com/ai2cm/gt4py.git
+cd gt4py
+git checkout 47021270f067aff314a0344637d090657f1460d8
+cd ../
+pip install -e ./gt4py
+
+if [ -d "dace" ]; then rm -rf dace; fi
+git clone https://github.com/spcl/dace.git
+cd dace
+git checkout 56c42b21b72b043408b725dd85d6a47ba62f451b
+git submodule update --init
+cd ../
+pip install -e ./dace
+
+set +e -x
+# TODO: this somehow does not work with 9 or 10
+module switch gcc gcc/8.3.0
+
+set -e -x
+if [ "$orchestration"  == "dace" ] ; then
+    echo "setting dace ochestration specific flags"
+    export FV3_DACEMODE=True
+    export DACE_execution_general_check_args=0
+    export DACE_frontend_dont_fuse_callbacks=1
+    export DACE_compiler_cpu_openmp_sections=0
+    export DACE_compiler_cuda_max_concurrent_streams=-1
+    export DACE_frontend_unroll_threshold=0
+    export DACE_compiler_unique_functions=none
+fi
--- a/fv3core/.jenkins/jenkins.sh
+++ b/fv3core/.jenkins/jenkins.sh
@@ -60,6 +60,8 @@ TOP_LEVEL_JENKINS_DIR=$JENKINS_DIR/../../.jenkins
 action="$1"
 backend="$input_backend"
 experiment="$3"
+orchestration=${4:-none}
+savepoint=${5:-All}
 
 # check presence of env directory
 pushd `dirname $0` > /dev/null
@@ -201,9 +203,22 @@ if [ ${python_env} == "virtualenv" ]; then
     fi
     export FV3_PATH="${JENKINS_DIR}/../"
     export TEST_DATA_RUN_LOC=${TEST_DATA_HOST}
+    if [ "$orchestration"  == "dace" ] ; then
+        echo "dace orchestration is turned on"
+        ${JENKINS_DIR}/install_dace.sh $orchestration
+    elif [ "$orchestration"  == "python" ] ; then
+        echo "no dace orchestration, but running with dace backend"
+        ${JENKINS_DIR}/install_dace.sh $orchestration
+    else
+        echo "no dace orchestration, not running dace backend"
+    fi
 fi
 
-run_command "${script} ${backend} ${experiment} " Job${action} ${scheduler_script}
+if [ "$orchestration"  == "none" ] ; then
+    run_command "${script} ${backend} ${experiment} " Job${action} ${scheduler_script}
+else
+    run_command "${script} ${backend} ${experiment} ${savepoint} ${orchestration}" Job${action} ${scheduler_script}
+fi
 
 if [ $? -ne 0 ] ; then
   exitError 1510 ${LINENO} "problem while executing script ${script}"
--- a/fv3core/Makefile
+++ b/fv3core/Makefile
@@ -11,6 +11,7 @@ PULL ?=True
 DEV ?=n
 NUM_RANKS ?=6
 VOLUMES ?=
+DACE_INSTALL ?=
 CONTAINER_ENGINE ?=docker
 RUN_FLAGS ?=--rm
 TEST_ARGS ?=
@@ -120,7 +121,7 @@ ifneq ($(findstring docker,$(CONTAINER_CMD)),)
 	$(MAKE) build
     endif
 endif
-	$(CONTAINER_CMD) bash -c "pip list && $(PYTEST_CMD)"
+	$(CONTAINER_CMD) bash -c "${DACE_INSTALL} pip list && $(PYTEST_CMD)"
 
 sync_test_data:
 	mkdir -p $(TEST_DATA_HOST) && gsutil -m rsync -r $(DATA_BUCKET) $(TEST_DATA_HOST)


fv3core/.jenkins/actions/run_regression_tests.sh
```diff
@@ -2,8 +2,17 @@
 set -e -x
```
mcgibbon, line 2: Can you base this PR off of the feature/dace-orchestrate-delnflux branch, since we only plan to run it in branches based off that branch (or main after it's been merged into main)?

fv3core/.jenkins/actions/run_regression_tests.sh
```diff
@@ -2,8 +2,17 @@
...
 BACKEND=$1
 EXPNAME=$2
+SAVEPOINT=${3:-All}
+ORCHESTRATION=${4:-none}
 XML_REPORT="sequential_test_results.xml"
-export TEST_ARGS="-v -s -rsx --backend=${BACKEND} "
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
mcgibbon, line 8: If we install `dace` as part of the base environment, you can avoid all of the changes here apart from the savepoint selection. We could avoid the savepoint selection as well if we instead used a hard-coded list of savepoint names in `conftest.py` that we want to test with dace, and updated those in our branch as new ones work. Has the added benefit of regression testing all dace-working code.

fv3core/.jenkins/jenkins.sh
```diff
@@ -201,9 +203,22 @@ if [ ${python_env} == "virtualenv" ]; then
     fi
     export FV3_PATH="${JENKINS_DIR}/../"
     export TEST_DATA_RUN_LOC=${TEST_DATA_HOST}
+    if [ "$orchestration"  == "dace" ] ; then
```
mcgibbon, line 206: Especially if this logic has to be copy-pasted into multiple bash scripts, I really think we should include dace as part of "the pace environment" always.

fv3core/.jenkins/install_dace.sh
```diff
@@ -0,0 +1,33 @@
...
+git clone https://github.com/spcl/dace.git
+cd dace
+git checkout 56c42b21b72b043408b725dd85d6a47ba62f451b
+git submodule update --init
+cd ../
+pip install -e ./dace
+
+set +e -x
+# TODO: this somehow does not work with 9 or 10
+module switch gcc gcc/8.3.0
```
mcgibbon, line 21: Should this change be propagated to the module load commands in buildenv for daint?

mcgibbon (changes requested):
I think we can avoid nearly if not all of the bash changes here. I'd ask that we include dace in "the environment" to avoid special pip installs for anything using it, and that the dace-specific hardcoded environment variable exports get moved to `dsl/pace/dsl/dace/__init__.py` where we already have one.

Optionally, you could also keep a list of DACE_COMPATIBLE_SAVEPOINTS in fv3core's conftest.py, and have only those savepoints get loaded when the environment variable `FV3_DACEMODE=True` is set.

If you can do all three of these changes, it should avoid most if not all of the changes you needed to make to the sh scripts here, instead putting code in python files and modifying the daint environment to be compatible with dace (I hope there's one gcc that works with everything).

fv3core/.jenkins/jenkins.sh
elynnwu: The problem here is not installing dace, the naming is confusing. What triggers the install of dace is through backend, specifically `gtc:dace` or `gtc:dace_gpu`. However, in `gtc:dace`, there're two modes: one without and one with dace orchestration. Basically if you want to have orchestrated `gtc:dace`, you need to set those environment flags.

fv3core/.jenkins/install_dace.sh
elynnwu: This is a temporary fix, confirmed with @twicki. This should be resolved after the conference paper, and we would like this to be consistent with everything else in buildev.

fv3core/.jenkins/jenkins.sh
```diff
@@ -60,6 +60,8 @@ TOP_LEVEL_JENKINS_DIR=$JENKINS_DIR/../../.jenkins
 action="$1"
 backend="$input_backend"
 experiment="$3"
+orchestration=${4:-none}
```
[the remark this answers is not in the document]
elynnwu: Oh I understand now. Yes, we should just have `FV3_DACEMODE` as an jenkins config.

fv3core/.jenkins/jenkins.sh
elynnwu: Never mind, I understand now. Including dace as part of install requirements.

fv3core/.jenkins/jenkins.sh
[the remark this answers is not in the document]
elynnwu: But there's just one problem - we still need to pass this environment flag into docker container when running jenkins gce.

dsl/pace/dsl/dace/__init__.py
```diff
@@ -5,3 +5,9 @@
     "DACE_CONFIG",
     os.path.join(os.path.dirname(os.path.realpath(__file__)), "dace.conf"),
 )
+os.environ.setdefault("DACE_execution_general_check_args", "0")
+os.environ.setdefault("DACE_frontend_dont_fuse_callbacks", "1")
+os.environ.setdefault("DACE_compiler_cpu_openmp_sections", "0")
+os.environ.setdefault("DACE_compiler_cuda_max_concurrent_streams", "-1")
+os.environ.setdefault("DACE_frontend_unroll_threshold", "0")
+os.environ.setdefault("DACE_compiler_unique_functions", "none")
```
elynnwu, line 13: We set these regardless of `FV3_DACEMODE` because it wouldn't matter when it's false.

fv3core/.jenkins/actions/run_regression_tests.sh
```diff
@@ -15,5 +15,9 @@ if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="srun" make tests savepoint_tests
 else
     export TEST_ARGS="${TEST_ARGS} --junitxml=/.jenkins/${XML_REPORT}"
-    VOLUMES="-v ${pwd}/.jenkins:/.jenkins" make tests savepoint_tests
+    if [[ ${FV3_DACEMODE} == "True" ]]; then
+        VOLUMES="-v ${SCRIPT_DIR}/../:/.jenkins" RUN_FLAGS="--rm -e FV3_DACEMODE=True" make savepoint_tests
+    else
+        VOLUMES="-v ${SCRIPT_DIR}/../:/.jenkins" make tests savepoint_tests
```
elynnwu, line 21: Actually, unless there's a more elegant way of doing this, we want to run only `savepoint_tests` when `FV3_DACEMODE=True`. I think we could keep RUN_FALGS here instead of appending in the Makefile, because the work flow would have been similar:
```
export FV3_DACEMODE=True
cd fv3core
.jenkins/jenkins.sh run_regression_tests gtc:dace c12_6ranks_standard
```

vs.

```
export FV3_DACEMODE=True
cd fv3core
make savepoint_tests "TEST_ARGS=--backend=gtc:dace"
```

dsl/pace/dsl/dace/__init__.py
mcgibbon, line 13:
```suggestion
# we set these regardless of FV3_DACEMODE because they have no effect without dace
os.environ.setdefault("DACE_execution_general_check_args", "0")
os.environ.setdefault("DACE_frontend_dont_fuse_callbacks", "1")
os.environ.setdefault("DACE_compiler_cpu_openmp_sections", "0")
os.environ.setdefault("DACE_compiler_cuda_max_concurrent_streams", "-1")
os.environ.setdefault("DACE_frontend_unroll_threshold", "0")
os.environ.setdefault("DACE_compiler_unique_functions", "none")
```

fv3core/.jenkins/actions/run_parallel_regression_tests.sh
```diff
@@ -16,5 +16,9 @@ export TEST_ARGS="${TEST_ARGS} --compute_grid"
 if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="" make savepoint_tests_mpi
 else
-    make savepoint_tests_mpi
+    if [[ ${FV3_DACEMODE} == "True" ]]; then
+        RUN_FLAGS="--rm -e FV3_DACEMODE=True" make savepoint_tests_mpi
```
mcgibbon, line 20: Can you put this logic in the Makefile instead, so that it happens automatically when we test as a user? I'd expect to be able to run `FV3_DACEMODE=True TEST_ARGS="--backend=\"gtc:gt:dace\" make savepoint_tests` (modulo using the correct backend name, if that's not it). If you want to automatically set the backend when `FV3_DACEMODE=True` to remove the TEST_ARGS bit, I wouldn't complain (but be on the lookout for FV3_DACEMODE=True being set when running other-backend tests on jenkins).

fv3core/.jenkins/actions/run_regression_tests.sh
```diff
@@ -12,8 +12,16 @@ make get_test_data
 
 if [ ${python_env} == "virtualenv" ]; then
     export TEST_ARGS="${TEST_ARGS} --junitxml=${JENKINS_DIR}/${XML_REPORT}"
-    CONTAINER_CMD="srun" make tests savepoint_tests
+    if [[ ${FV3_DACEMODE} == "True" ]]; then
+        CONTAINER_CMD="srun" make savepoint_tests
+    else
+        CONTAINER_CMD="srun" make tests savepoint_tests
```
mcgibbon, line 18: Should we move `make tests` onto its own action, or even delete it entirely to run on CircleCI only? I'm a little frustrated when I can't get jenkins savepoint results because the main tests failed. Maybe a question for a team meeting.

fv3core/.jenkins/actions/run_regression_tests.sh
```diff
@@ -12,8 +12,16 @@ make get_test_data
 
 if [ ${python_env} == "virtualenv" ]; then
     export TEST_ARGS="${TEST_ARGS} --junitxml=${JENKINS_DIR}/${XML_REPORT}"
-    CONTAINER_CMD="srun" make tests savepoint_tests
+    if [[ ${FV3_DACEMODE} == "True" ]]; then
+        CONTAINER_CMD="srun" make savepoint_tests
```
mcgibbon, line 16: If you make more edits to this file, can you move some logic out of the virtualenv if-statement by having it control exporting environment variables only, and then having one line (or two, if you need the dacemode if-statement) calling `make`? Two `make` lines is maybe OK but four is getting up there.

fv3core/.jenkins/actions/run_regression_tests.sh
mcgibbon: I normally run tests using make inside the fv3core directory, and ideally would like to continue doing this. Having the logic inside the Makefile means the same logic gets re-used for make targets and jenkins actions.

I usually expect calling `jenkins.sh` to fail when running on my local machine. For example, `./.jenkins/jenkins.sh run_regression_tests numpy c12_6ranks_standard` crashes with a "slave is not defined" error, for me.

I think it's fine if `FV3_DACEMODE=True make test` is expected to fail, your if-conditionals in the actions scripts avoid doing this.

mcgibbon: You can still keep the if-statement to toggle whether we're running the `main` tests, if that stays here.

fv3core/.jenkins/jenkins.sh
```diff
@@ -201,6 +201,12 @@ if [ ${python_env} == "virtualenv" ]; then
     fi
     export FV3_PATH="${JENKINS_DIR}/../"
     export TEST_DATA_RUN_LOC=${TEST_DATA_HOST}
+    if [[ "$backend"  == *"dace"* ]] ; then
+        echo "temporary fix for dace on daint..."
```
mcgibbon, line 205:
```suggestion
        echo "temporary fix for dace on daint, replacing gcc with 8.3.0..."
```

fv3core/.jenkins/actions/run_regression_tests.sh
mcgibbon: If you can, please append (instead of set default for) `-e FV3_DACEMODE=True`, so we can still pass manual RUN_FLAGS without copy-pasting or remembering this flag.

mcgibbon (commented):
Only change I still would really like to see is moving the FV3_DACEMODE environment variable declaration into the Makefile. We can chat more if there are reasons not to do this.

fv3core/.jenkins/actions/run_parallel_regression_tests.sh
elynnwu: Moved to Makefile.

fv3core/.jenkins/actions/run_regression_tests.sh
elynnwu: Appending `RUN_FLAGS` and passed in Makefile.

fv3core/.jenkins/actions/run_regression_tests.sh
elynnwu: I'm keeping this for now because passing `CONTAINER_CMD` and `VOLUMES` to makefile is quite difficult unless it's explicitly passed like `CONTAINER_CMD="srun" make savepoint_tests`. But then you need to have more if statements to be able to run for both daint and gce. Since there's a chance that we don't even run `make tests` (comments above), maybe it's not an issue here.

mcgibbon: I thought Makefiles read environment variables? e.g. if you `export CONTAINER_CMD="srun"` then `make savepoint_tests`, it's the same as if you `CONTAINER_CMD="srun" make savepoint_tests`, apart from that after the command is run the environment variable will still be set (which shouldn't break anything here).

elynnwu: You're right, I must have missed `export` because it didn't work when I tried yesterday. Updating this.

PR #149, opened 2022-04-01
Format source when using numpy backend

jdahm:
## Purpose

In the tests, format the generated numpy source when using the gtc:numpy debug backend.

## Code changes:

- Sets `format_source = True` for tests when using backend with numpy in their name.


rheacangeo (approved):

mcgibbon:
launch jenkins

PR #153, opened 2022-04-08
Add CachingCommReader and CachingCommWriter

mcgibbon:
## Purpose

We would like to be able to debug our code while running in serial, which is not possible due to inter-process communciation. While we can already test with invalid output data using mocks that ignore communication, we need another solution to keep correct values while still running in serial.

This PR adds two classes mocking the mpi4py Comm object to enable writing communication to disk, then running in a single process while re-playing that communication. It also formalizes the Comm type with an abstract base class, and brings the NullComm from fv3core into pace.util.

## Public code changes:

- Added `CachingCommWriter` which wraps a `Comm` object and can be serialized to a file-like object with a `.dump` method
- Added `CachingCommReader` which can be loaded from the dump output of `CachingCommWriter` and replays its communication in the order it occurred.
- `NullComm` is now public api in `pace-util`
- Deleted deprecated `finish_vector_halo_update` method from `CubedSphereCommunicator`

## Private code changes:

- Added Comm and Request abstract base classes to ensure we implement all required methods from mpi4py
- Moved vector-interface-synchronization functionality onto new VectorInterfaceHaloUpdater class **(note for reviewers: I could revert this, it's not needed since I decided to mock `Comm` instead of `HaloUpdater`, but this should be a good refactor anyways)**

## Checklist


pace-util/pace/util/caching_comm.py
```diff
@@ -0,0 +1,201 @@
...
+        return self._data.get_bcast()
+
+    def barrier(self):
+        pass
+
+    def Scatter(self, sendbuf, recvbuf, root=0, **kwargs):
+        recvbuf[:] = self._data.get_buffer()
+
+    def Gather(self, sendbuf, recvbuf, root=0, **kwargs):
+        recvbuf[:] = self._data.get_buffer()
```
mcgibbon, line 115: get_buffer will sometimes return None for this, because scatter/gather don't always receive data on every rank. Have to edit to fix this, should probably add a test.

oelbert (changes requested):
One issue with the CachingCommReader we'd discussed, and I think you want to update the History.md for pace-util, but otherwise it looks good. Thanks for cleaning up the halo updaters while you were here!

jdahm:
launch jenkins

mcgibbon:
launch jenkins

oelbert (approved):
Great! Thanks for putting this together

PR #170, opened 2022-04-18
util/v0.8.0 + zarr monitor: allow transposed diagnostics to be written

brianhenn:
## Purpose

The zarr monitor currently ascertains the metadata of the variable to be written from the first chunk (in rank or time) that it receives, and assumes all future chunks will have the same metadata including dimension order. Thus it fails if an upstream process transposes a subsequent chunk (e.g., https://github.com/ai2cm/fv3net/issues/1767) , but it has the needed information to handle this case. This PR allows for dimension order changes in diagnostic chunks so long as the other metadata stays the same across chunks. 

Also updates `pace-util` to `v0.8.0`.

## Code changes:

- `ZarrMonitor.store` now allow for transposing inputs to match previously stored dimension order, if inputs have same set of dimensions but in different order.

## Checklist

- [X] Docstrings and type hints are added to new and updated routines, as appropriate
- [X] All relevant documentation has been updated or added (e.g. README, CONTRIBUTING docs)
- [X] For each public change and fix in `pace-util`, HISTORY has been updated
- [X] Unit tests are added or updated for non-stencil code changes

### DIFF
--- a/pace-util/pace/util/zarr_monitor.py
+++ b/pace-util/pace/util/zarr_monitor.py
@@ -121,8 +121,9 @@ class ZarrMonitor:
         Requires the state contain the same quantities with the same metadata as the
         first time this is called. Dimension order metadata may change between calls
         so long as the set of dimsensions is the same. Quantities are stored with
-        dimensions [time, rank] followed by the dimensions included in any one state
-        snapshot. The one exception is "time" which is stored with dimensions [time].
+        dimensions [time, rank] followed by the dimensions included in the first
+        state snapshot. The one exception is "time" which is stored with dimensions
+        [time].
         """
         self._ensure_writers_are_consistent(state)
         for name, quantity in sorted(state.items(), key=lambda x: x[0]):
@@ -266,12 +267,6 @@ class _ZarrVariableWriter:
             "_ARRAY_DIMENSIONS": list(self._PREPEND_DIMS + quantity.dims),
             **quantity.attrs,
         }
-
-    def _order_dims(self, quantity):
-        if self.array is None:
-            return quantity
-        else:
-            dim_order = self.array.attrs["_ARRAY_DIMENSIONS"]
 
     def _ensure_compatible_attrs(self, new_quantity):
         new_attrs = self._get_attrs(new_quantity)
--- a/pace-util/pace/util/zarr_monitor.py
+++ b/pace-util/pace/util/zarr_monitor.py
@@ -119,9 +119,10 @@ class ZarrMonitor:
         """Append the model state dictionary to the zarr store.
 
         Requires the state contain the same quantities with the same metadata as the
-        first time this is called. Quantities are stored with dimensions [time, rank]
-        followed by the dimensions included in any one state snapshot. The one exception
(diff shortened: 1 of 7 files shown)


pace-util/pace/util/zarr_monitor.py
```diff
@@ -119,9 +119,10 @@ def store(self, state: dict) -> None:
         """Append the model state dictionary to the zarr store.
 
         Requires the state contain the same quantities with the same metadata as the
-        first time this is called. Quantities are stored with dimensions [time, rank]
-        followed by the dimensions included in any one state snapshot. The one exception
-        is "time" which is stored with dimensions [time].
+        first time this is called. Dimension order metadata may change between calls
+        so long as the set of dimsensions is the same. Quantities are stored with
+        dimensions [time, rank] followed by the dimensions included in any one state
+        snapshot. The one exception is "time" which is stored with dimensions [time].
```
mcgibbon, line 125: Have to make this distinction since your edit.
```suggestion
        dimensions [time, rank] followed by the dimensions included in the first state
        snapshot. The one exception is "time" which is stored with dimensions [time].
```

pace-util/pace/util/zarr_monitor.py
```diff
@@ -243,6 +267,12 @@ def _get_attrs(self, quantity):
             **quantity.attrs,
         }
 
+    def _order_dims(self, quantity):
+        if self.array is None:
+            return quantity
+        else:
+            dim_order = self.array.attrs["_ARRAY_DIMENSIONS"]
```
mcgibbon, line 274: Oh cool, didn't know you can get them like this. Is this public API in zarr?

pace-util/tests/test_zarr_monitor.py
```diff
@@ -394,3 +398,72 @@ def test_monitor_file_store_inconsistent_calendars(
...
+            mpi_comm=DummyComm(
+                rank=rank, total_ranks=total_ranks, buffer_dict=shared_buffer
+            ),
+        )
+        if rank % 2 == 0:
+            diag_to_store = _transpose(diag)
+        else:
+            diag_to_store = _transpose(
+                diag, dims_2d=[X_DIMS, Y_DIMS], dims_3d=[X_DIMS, Y_DIMS, Z_DIMS]
+            )
```
mcgibbon, line 448: Could you please make the arguments required and move the currently-default arguments here and into the other test to be more explicit (at the cost of repetition)? It would make it clearer to be able to read that the dimension transposing is different each time, also that they do get transposed in both cases (as opposed to `_transpose` doing nothing if passed no args).

mcgibbon (approved):
Some minor doc edits, one very minor nit in the tests. Thanks!

pace-util/pace/util/zarr_monitor.py
brianhenn: Ha whoops this function was a leftover from a previous approach, will delete. But to answer your question, `zarr.core.Array.attrs` is public API (as a MutableMapping), so I think this is fine, and I do use this to check and match dim orders above.

mcgibbon:
launch jenkins

PR #172, opened 2022-04-21
disable driver diagnostics if no path given

mcgibbon:
## Purpose

Currently it is not possible to avoid creating a Zarr store with at least the grid diagnostics when running the model. However, using a zarr store makes it difficult to port cached communications between systems, as non-root ranks will crash if they receive a zarr group describing a store which does not exist.

This PR modifies the behavior of pace.driver.run so that if no path is given for the diagnostics (previously a path was required), no diagnostics will be stored.

## Code changes:

- DiagnosticsConfig's path argument is now optional, if not given then no diagnostics are stored.

## Checklist

### DIFF
--- a/driver/pace/driver/diagnostics.py
+++ b/driver/pace/driver/diagnostics.py
@@ -1,6 +1,7 @@
+import abc
 import dataclasses
 from datetime import datetime
-from typing import List
+from typing import List, Optional
 
 import zarr.storage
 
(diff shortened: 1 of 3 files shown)


elynnwu (commented):
Just to be clear, it looks like if you don't specify anything in path, it will default to `NullDiagnostics`. So if you have names specified and no path, nothing will be saved? Also, can you have `path: None` in the yaml file? I think it would be a good idea to have an example yaml or README on this.

mcgibbon:
> Just to be clear, it looks like if you don't specify anything in path, it will default to `NullDiagnostics`. So if you have names specified and no path, nothing will be saved? Also, can you have `path: None` in the yaml file? I think it would be a good idea to have an example yaml or README on this.

Yes, I tried to make that clear from the docstring for the diagnostics configuration class. Should I edit it?

elynnwu (approved):
lgtm!

PR #174, opened 2022-04-26
Add Jupyter notebook example running driver with write-read

mcgibbon:
## Purpose

This PR adds a Jupyter notebook example running the driver in a write-comm mode, then debugging it in read mode.

It also contains bug fixes and features used to implement that example.

## Code changes:

- Added notebook example using driver to write data in MPI and then read it in serial under driver/examples/notebooks
- Added defaults for DiagnosticsConfig, PerformanceConfig, and CommConfig
- Added `driver.step(timestep)` public method to Driver
- Added call to gc.collect() at end of driver execution which is necessary to avoid MPI error messages 
- When initialized from a storage, Quantity.data is now set as storage.data instead of np.asarray(storage.data), avoiding the potential for a data copy/mismatch between these two objects

## Checklist


pace-util/tests/quantity/test_storage.py
```diff
@@ -133,6 +133,22 @@ def test_modifying_storage_modifies_data(quantity):
     storage = quantity.storage
     quantity.data[:] = 5
     assert quantity.np.all(quantity.np.asarray(storage) == 5)
+    assert quantity.data.data == quantity.storage.data.data
+
+
+@pytest.mark.parametrize("backend", ["gt4py_numpy"], indirect=True)
+def test_modifying_storage_modifies_data_when_initialized_from_storage(quantity):
```
mcgibbon, line 140: This test ended up passing on main, but I think it's still worth keeping.

mcgibbon:
I didn't add a unit test for `step` (e.g. that it behaves the same as step_all when used), but I did try to make sure all the logic is pulled into it so that `step_all` is clearly just calling `step` repeatedly and doing nothing else.

driver/pace/driver/diagnostics.py
```diff
@@ -17,7 +17,7 @@
 
 class Diagnostics(abc.ABC):
     @abc.abstractmethod
-    def store(self, time: datetime, state: DriverState):
+    def store(self, time: Union[datetime, timedelta], state: DriverState):
```
elynnwu, line 20: Why do we need to store timedelta?

driver/pace/driver/driver.py
```diff
@@ -238,28 +243,25 @@ def __init__(
         log_subtile_location(
             partitioner=communicator.partitioner.tile, rank=communicator.rank
         )
+        self.diagnostics.store_grid(
+            grid_data=self.state.grid_data,
+            metadata=self.state.dycore_state.ps.metadata,
+        )
```
elynnwu, line 249: Moving this to init makes a lot of sense!

driver/pace/driver/driver.py
```diff
@@ -70,9 +70,13 @@ class DriverConfig:
...
-    diagnostics_config: diagnostics.DiagnosticsConfig
-    performance_config: PerformanceConfig
-    comm_config: CommConfig
+    diagnostics_config: diagnostics.DiagnosticsConfig = dataclasses.field(
+        default_factory=diagnostics.DiagnosticsConfig
+    )
+    performance_config: PerformanceConfig = dataclasses.field(
+        default_factory=PerformanceConfig
+    )
+    comm_config: CommConfig = dataclasses.field(default_factory=CommConfig)
```
elynnwu, line 79: Is the idea of having default type for these to make demo simple? One argument is that not having default type ensures users to specify everything they're supposed to in the namelist.

driver/pace/driver/driver.py
```diff
@@ -185,49 +189,50 @@ def __init__(
...
+        self.time = self.config.start_time
         self.comm_config = config.comm_config
         self.comm = config.comm_config.get_comm()
         self.performance_config = self.config.performance_config
         with self.performance_config.total_timer.clock("initialization"):
             communicator = pace.util.CubedSphereCommunicator.from_layout(
                 comm=self.comm, layout=self.config.layout
             )
-            quantity_factory, stencil_factory = _setup_factories(
+            self.quantity_factory, self.stencil_factory = _setup_factories(
```
elynnwu, line 200: I don't think these two are used in call time? Is it more of a consistent style choice?

driver/pace/driver/driver.py
mcgibbon: Not really - CommConfig and the others here already have defaults set for all of their options, so the idea was that if you can and would ever set `comm_config: {}` it's probably better to be able to remove this from the yaml entirely if you want the default.

If we wanted this to be required, we should make it so that there is no default communication type ("mpi" currently). But I feel like it makes sense as a default behavior.

driver/pace/driver/driver.py
mcgibbon: I wanted these to become public API so it's easier to interact with a `driver` class externally, e.g. to initialize a d2a2c object that matches the settings of the driver and pass the driver's state into it (which I do in the demo notebook).

elynnwu (approved):

PR #178, opened 2022-05-06
Fix Overrides

rheacangeo:
fix test overrides files that refer to the numpy backend, when it should be gtc:numpy, or not specified (because if it is needed for gtc:numpy, it almost certainly is needed for the other backends)


rheacangeo:
launch jenkins

mcgibbon:
launch jenkins

FlorianDeconinck (approved):
LGTM

PR #185, opened 2022-05-12
Add restart capability to pace.driver

elynnwu:
## Purpose

We add the capability to output pace `dycore`, `physics`, and `tendency` state and restart from these states later. 

## Code changes:

- `save_restart` is now an option in driver yaml to indicate outputting the last time step
- `driver.py`: during cleanup, we save dycore, physics, and tendency state in a RESTART folder as well as the driver config to be used during restart (default to RESTART/restart.yaml)
- `RestartConfig` is now implemented and uses `_restart_driver_state`
  - when loading from restart, we calculate `grid_data`, `damping_coefficients`, `driver_grid_data` based on the input driver config
- `driver/state.py`: `DriverState` can now save and load from RESTART, logics to map restart files to their pace state are under `_restart_driver_state` and are used in both `RestartConfig` and `DriverState.load_state_from_restart`
- class property `xr_dataset` is added to `DycoreState`, `PhysicsState`, and `TendencyState`

## Requirements changes:

- circleci driver test now requires additional packages to handle netcdf and zarr files

## Infrastructure changes:

- Added additional tests for restart
  - mainly, we make sure the following produce the same results: 
    1. run one step, output restart, load restart and run one more step and save as diagnostics
    2. run two steps directly from initialization

## Checklist


driver/Makefile
```diff
@@ -1,8 +1,25 @@
 MPIRUN_CALL ?= mpirun -n 6
 
 test:
-	pytest tests
+	pytest tests --ignore=tests/test_restart.py
```
mcgibbon, line 4: Is there any way to avoid this? e.g. can we have a `tests/main` and `tests/mpi` separation for the tests that need MPI?

driver/Makefile
```diff
@@ -1,8 +1,25 @@
...
+
+test_restart:
+	cp examples/configs/baroclinic_c12_write_restart.yaml baroclinic_c12_write_restart.yaml
+	cp examples/configs/baroclinic_c12_write_restart.yaml baroclinic_c12_run_two_steps.yaml
+	sed -i 's/seconds: 225/seconds: 450/' baroclinic_c12_run_two_steps.yaml
+	sed -i 's/save_restart: true/save_restart: false/' baroclinic_c12_run_two_steps.yaml
+	sed -i 's/path: "output.zarr"/path: "run_two_steps_output.zarr"/' baroclinic_c12_run_two_steps.yaml
+	$(MPIRUN_CALL) -n 6 python3 -m pace.driver.run baroclinic_c12_write_restart.yaml
+	$(MPIRUN_CALL) -n 6 python3 -m pace.driver.run RESTART/restart.yaml
+	$(MPIRUN_CALL) -n 6 python3 -m pace.driver.run baroclinic_c12_run_two_steps.yaml
```
mcgibbon, line 18: Can these calls be done inside `test_restart`? Would make them a little easier to run the test and to diagnose what's going on if we run test_restart manually and the files are missing.

driver/examples/configs/baroclinic_c12_write_restart.yaml
```diff
@@ -0,0 +1,90 @@
...
+  experiment_name: c12_baroclinic
+nx_tile: 12
+nz: 79
+dt_atmos: 225
+minutes: 0
+seconds: 225
+layout:
+  - 1
+  - 1
+save_restart: true
```
mcgibbon, line 20: Do we want/need to configure how often restarts are saved? It could slow down the simulation a lot to do this every timestep.

driver/pace/driver/initialization.py
```diff
@@ -125,26 +122,18 @@ class RestartConfig(Initializer):
     Configuration for restart initialization.
     """
 
-    path: str
-
-    @property
-    def start_time(self) -> datetime:
-        return datetime(2000, 1, 1)
+    path: str = "."
+    start_time: datetime = datetime(2000, 1, 1)
```
mcgibbon, line 126: We can ignore this for now, but we should probably attach the model time to DriverState in the future and save it in the netcdfs.

driver/pace/driver/registry.py
```diff
@@ -115,7 +115,8 @@ def from_dict(self, config: dict) -> T:
         It can also have a "config" key, which is a dict used to initialize the
         dataclass. By default this is an empty dict.
         """
-        config.setdefault("config", {})
+        if "config" not in config:
+            config.setdefault("config", {})
```
mcgibbon, line 119: `setdefault` only modifies the dictionary if the key does not exist, the if statement would be redundant.
```suggestion
        config.setdefault("config", {})
```

driver/pace/driver/state.py
```diff
@@ -58,3 +81,108 @@ class DriverState:
     grid_data: pace.util.grid.GridData
     damping_coefficients: pace.util.grid.DampingCoefficients
     driver_grid_data: pace.util.grid.DriverGridData
+
+    @classmethod
+    def load_state_from_restart(
```
mcgibbon, line 86: It would make sense to include a unit test that if you save a DriverState to restart and then load it, all data is the same (by iterating over every dataclass field and comparing). Would be helpful for diagnosing any issues that come up, e.g. if we add something to DriverState but we don't add it to the serialization.

driver/pace/driver/state.py
```diff
@@ -58,3 +81,108 @@ class DriverState:
...
+        )
+        self.physics_state.xr_dataset.to_netcdf(
+            f"{restart_path}/restart_physics_state_{current_rank}.nc"
+        )
+        self.tendency_state.xr_dataset.to_netcdf(
+            f"{restart_path}/restart_tendency_state_{current_rank}.nc"
+        )
+
+
+def _read_restart_files(
```
mcgibbon, line 129: Maybe `overwrite_state_from_restart` to make it clear we're mutating `state`?

driver/requirements.txt
```diff
@@ -3,3 +3,7 @@ pytest-subtests
 pytest-cov
 dacite
 mpi4py
+numpy
+netCDF4
+xarray
+zarr
```
mcgibbon, line 9: Do any of these need to be added to driver's `setup.py`?

fv3gfs-physics/fv3gfs/physics/physics_state.py
```diff
@@ -1,6 +1,9 @@
 from dataclasses import InitVar, dataclass, field, fields
 from typing import List, Optional
 
+import gt4py.gtscript as gtscript
+import xarray as xr
```
mcgibbon, line 5: Will need to add xarray to `fv3gfs-physics/setup.py`.

fv3core/fv3core/initialization/dycore_state.py
```diff
@@ -1,6 +1,8 @@
 from dataclasses import dataclass, field, fields
 from typing import Any, Mapping
 
+import xarray as xr
```
mcgibbon, line 4: Will need to add xarray to `fv3core/setup.py`.

driver/tests/test_restart.py
```diff
@@ -0,0 +1,72 @@
...
+        layout=(1, 1),
+        initialization=unittest.mock.MagicMock(),
+        performance_config=unittest.mock.MagicMock(),
+        comm_config=NullCommConfig((1, 1)),
+        diagnostics_config=unittest.mock.MagicMock(),
+        dycore_config=unittest.mock.MagicMock(nwat=6),
+        physics_config=unittest.mock.MagicMock(),
+        save_restart=True,
+    )
+    assert driver_config.save_restart is True
```
mcgibbon, line 34: I think this is too trivial of a test to include for how complex it is (lots of mocks needed for DriverConfig), here we're basically testing that dataclasses work.

If you want to test the default behavior is to save restarts (or that it's not to save restarts) that would make sense but you'll want to remove it from the keyword arguments passed if so.

driver/examples/configs/baroclinic_c12_write_restart.yaml
elynnwu: Currently, it is called as part of driver cleanup, so only once. Because of the try/finally logic, I believe this would've saved the last time step before model crashes as well. 

I think for our use case, we just want to save the last time step. We could add logic to do restart frequency later if we have a use for that.

driver/Makefile
elynnwu: Moved inside `test_restart`

driver/pace/driver/registry.py
elynnwu: Didn't realize that, thanks!

driver/requirements.txt
elynnwu: Yes, added

driver/tests/test_restart.py
elynnwu: Changed to testing default value (should be False)

driver/pace/driver/state.py
elynnwu: Added this test under `test_driver_state_load_restart`

driver/tests/mpi/test_restart.py
```diff
@@ -0,0 +1,111 @@
...
+        comm_config=NullCommConfig((1, 1)),
+        diagnostics_config=unittest.mock.MagicMock(),
+        dycore_config=unittest.mock.MagicMock(nwat=6),
+        physics_config=unittest.mock.MagicMock(),
+    )
+    assert driver_config.save_restart is False
+
+
+def test_restart_results():
+    subprocess.call("tests/mpi/run_save_and_load_restart.sh")
```
mcgibbon, line 54: Can you add a `try: finally` block in this test that checks for and deletes the RESTART folder to clean up the test?

driver/tests/mpi/test_restart.py
```diff
@@ -0,0 +1,111 @@
...
+        "qsnow",
+        "qgraupel",
+    ]:
+        np.testing.assert_allclose(
+            restart[var].isel(time=0).values, regular[var].isel(time=-1).values
+        )
+
+
+@pytest.mark.last
+def test_driver_state_load_restart():
```
mcgibbon, line 82: It's pretty finnicky to rely on artifacts from a previous test to be used in a later test. Can you combine this with the test that runs the model? It can be used as a helper function if that's easier.

driver/examples/configs/baroclinic_c12_write_restart.yaml
mcgibbon: Ah for now saving at the end should be good enough. There's decent reason to want to save only at the end of timesteps periodically, since an exception in the middle of a timestep could give an inconsistent model state, but we don't need to worry about that for now.

jdahm:
launch jenkins

driver/tests/mpi/test_restart.py
elynnwu: Merged to one

.gitignore
```diff
@@ -6,6 +6,7 @@ results.xml
 driver/examples/*.json
 driver/examples/**/*.json
 driver/examples/comm
+20*-*-*-*-*-*.json
```
mcgibbon, line 9: Not for your PR, but we may want to reduce side-effect files of running the model or at least make the location these go to configurable.

driver/Makefile
```diff
@@ -1,8 +1,15 @@
 MPIRUN_CALL ?= mpirun -n 6
 
 test:
-	pytest tests
+	pytest tests/main
 
 test_mpi:
+	MPIRUN_CALL="$(MPIRUN_CALL)" pytest tests/mpi
```
mcgibbon, line 7: This is what I asked for and is I think the best we can do, but if others run into problems and ask about them be aware that `fv3core/tests/mpi` has to be run with `mpirun -n 6 pytest` whereas `driver/tests/mpi` has to be run with just `pytest`. May cause confusion, but should be fine since we have `make test_mpi` targets in both places.

driver/pace/driver/driver.py
```diff
@@ -55,6 +57,7 @@ class DriverConfig:
             including coupling code between the dycore and physics, as well as
             dry static adjustment. This is a development flag and will be removed
             in a later commit.
+        save_restart: whether to save the state output as restart files
```
mcgibbon, line 60:
```suggestion
        save_restart: whether to save the state output as restart files at
            cleanup time
```

driver/pace/driver/initialization.py
```diff
@@ -77,10 +77,7 @@ class BaroclinicConfig(Initializer):
     Configuration for baroclinic initialization.
     """
 
-    @property
-    def start_time(self) -> datetime:
-        # TODO: instead of arbitrary start time, enable use of timedeltas
-        return datetime(2000, 1, 1)
+    start_time: datetime = datetime(2000, 1, 1)
```
mcgibbon, line 80: No change needed, but for my information can `datetime` be specified in yaml? I excluded it before because I thought it's either not possible or a pain to do.

driver/pace/driver/initialization.py
```diff
@@ -125,26 +122,18 @@ class RestartConfig(Initializer):
...
     ) -> DriverState:
-        metric_terms = pace.util.grid.MetricTerms(
-            quantity_factory=quantity_factory, communicator=communicator
-        )
-        state = pace.util.open_restart(
-            dirname=self.path,
-            communicator=communicator,
-            quantity_factory=quantity_factory,
+        state = _restart_driver_state(
+            self.path, str(communicator.rank), quantity_factory, communicator
```
mcgibbon, line 134: Optional: change the argument type to `int` and cast to `str` inside the `_restart_driver_state` routine.

driver/pace/driver/state.py
mcgibbon: I don't see a test by that name, and the one test I do see calling `load_state_from_restart` doesn't test this (it compares the loaded state to the disk state, not to the before-saving state, so the conversion from before-saving to what's on disk is not tested).

driver/pace/driver/initialization.py
elynnwu: It worked if the format is correct, I have driver config writes out the restart yaml, and the format is this: `2000-01-01 00:07:30`. Then when it reads, it's able to parse that as `datetime`.

mcgibbon: Wow! Cool!

driver/tests/mpi/test_restart.py
```diff
@@ -0,0 +1,188 @@
...
+                    restart_physics[var].values,
+                )
+            else:
+                if var in restart_physics.keys():
+                    raise KeyError(
+                        f"{var} is not a storage and \
+                            should not be in physics restart file"
+                    )
+    finally:
+        subprocess.Popen(["make", "clean"])
```
mcgibbon, line 188: this looks like it could be dangerous - where does this run? is it always run in `driver/tests/mpi`?

I think you can add an argument to say where this command gets run, and pin it against `os.path.dirname(os.path.abspath(__file__))`.

driver/tests/mpi/test_restart.py
```diff
@@ -0,0 +1,188 @@
...
+            ):
+                np.testing.assert_allclose(
+                    driver_state.physics_state.__dict__[var].data,
+                    restart_physics[var].values,
+                )
+    finally:
+        subprocess.Popen(["make", "clean"])
+
+
+def test_restart_save_to_disk():
```
mcgibbon, line 125: is this an MPI test or should it go in `driver/tests/main`?

driver/tests/mpi/test_restart.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+        init = BaroclinicConfig()
+        driver_state = init.get_driver_state(
+            quantity_factory=quantity_factory, communicator=communicator
+        )
+        driver_state.save_state(mpi_comm)
+
+        restart_dycore = xr.open_dataset(
+            f"RESTART/restart_dycore_state_{mpi_comm.rank}.nc"
```
mcgibbon, line 154: this will only compare rank 0, is that intended?

driver/tests/mpi/test_restart.py
elynnwu: This will always run in `driver` because that's where the Makefile is. Alternatively, we can remove only RESTART folder and it might be less dangerous.

driver/tests/mpi/test_restart.py
elynnwu: Moved to main.

driver/tests/mpi/test_restart.py
elynnwu: Yes since it runs serially I think it's sufficient if we just test rank 0.

driver/tests/mpi/test_restart.py
mcgibbon: What if I'm in the top level and run `pytest driver/tests/mpi`? I do this kind of command pretty regularly, especially if I'm developing more than one package.

driver/tests/main/test_restart_serial.py
```diff
@@ -0,0 +1,113 @@
...
+                    restart_physics[var].values,
+                )
+            else:
+                if var in restart_physics.keys():
+                    raise KeyError(
+                        f"{var} is not a storage and \
+                            should not be in physics restart file"
+                    )
+    finally:
+        subprocess.Popen(["make", "clean"])
```
mcgibbon, line 113: Please change this to something like `shutil.rmdir("RESTART")`. In the future if we can make the restart output path configurable, we could tell it to write to a tmpdir.

mcgibbon (approved):
One change then LGTM! Thanks for all the iteration on this!

PR #186, opened 2022-05-16
Remove future_stencil

jdahm:
## Purpose

Removes the `FutureStencil` class and `future_stencil` entry-point function.

## Code changes:

- Removes the `FutureStencil` driver for gt4py stencils.
- Removes its tests.


dsl/pace/dsl/stencil.py
```diff
@@ -298,21 +296,10 @@ def __init__(
...
-        # Enable distributed compilation if running in parallel
-        if MPI is not None and MPI.COMM_WORLD.Get_size() > 1:
-            stencil_function = future_stencil.future_stencil
-            stencil_kwargs["wrapper"] = self
-        else:
-            # future stencil provides this information and
-            # we want to be consistent with the naming whether we are
-            # running in parallel or not (so we use the same cache)
-            stencil_kwargs["name"] = func.__module__ + "." + func.__name__
+        # stencil_kwargs["name"] = func.__module__ + "." + func.__name__
```
elynnwu, line 299: I think you want to uncomment this

elynnwu (approved):

dsl/pace/dsl/stencil.py
```diff
@@ -224,6 +229,7 @@ def __init__(
         )
         numpy_stencil_config = StencilConfig(
             backend="gtc:numpy",
+            cache_dir=".gt_cache",
```
elynnwu, line 232: Wouldn't this have to match with the rank as well? This is supposed to check if numpy give you the same answer.

driver/examples/configs/baroclinic_c12.yaml
```diff
@@ -1,5 +1,6 @@
 stencil_config:
   backend: gtc:numpy
+  cache_dir: .gt_cache_{rank}
```
elynnwu, line 3: Do we expect the user to change this?

dsl/pace/dsl/stencil.py
```diff
@@ -804,14 +795,20 @@ def get_quantity_halo_spec(
...
             grid_indexing: configuration for domain and halo indexing`
         """
         self.config: StencilConfig = config
         self.grid_indexing: GridIndexing = grid_indexing
+        self.comm: Comm = comm
+
+        # Replace {rank} in self.config.cache_dir with the communicator's rank
+        self.config.cache_dir = self.config.cache_dir.replace(
+            "{rank}", str(comm.Get_rank())
+        )
```
elynnwu, line 811: Is the idea here to allow 6 ranks to use one single cache as well? unless specified by the namelist to have 6 separate caches?

dsl/pace/dsl/stencil.py
jdahm: Ah, right. This should probably match the one in the `stencil_config` passed for consistency!

dsl/pace/dsl/stencil.py
jdahm: Yeah, the user can basically select programmatically, or through the driver yaml config.

dsl/pace/dsl/stencil.py
jdahm: I checked and I think the default is different, so yeah I'll change it back. Thanks!

jdahm:
It turns out that this did not jive well with the way `dace` requires it done for whole-program optimization.

PR #190, opened 2022-05-18
Add NullCheckpointer called for C_SW and D_SW

mcgibbon:
## Purpose

We are adding a checkpointer system to be able to test multiple savepoints in a single execution. This PR adds an abstract base class with a "no code" implementation so we can start adding calls to the checkpointer within our code.

## Code changes:

- Added Checkpointer and NullCheckpointer classes to pace-util
- Added `checkpointer` as arguments to AcousticDynamics and DynamicalCore, with a default NullCheckpointer value

## Checklist

### DIFF
--- a/fv3core/fv3core/stencils/dyn_core.py
+++ b/fv3core/fv3core/stencils/dyn_core.py
@@ -1,4 +1,4 @@
-from typing import Dict, Sequence, Union
+from typing import Dict, Optional, Sequence, Union
 
 from gt4py.gtscript import (
     __INLINED,
@@ -327,6 +327,7 @@ class AcousticDynamics:
         config: AcousticDynamicsConfig,
         pfull: FloatFieldK,
         phis: FloatFieldIJ,
+        checkpointer: Optional[pace.util.Checkpointer] = None,
     ):
         """
         Args:
@@ -341,6 +342,8 @@ class AcousticDynamics:
             pfull: atmospheric Eulerian grid reference pressure (Pa)
             phis: surface geopotential height
         """
+        self.call_checkpointer = checkpointer is not None
+        self.checkpointer = checkpointer
         grid_indexing = stencil_factory.grid_indexing
         self.comm = comm
         self.config = config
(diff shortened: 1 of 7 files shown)


elynnwu (commented):
Are we concerned about accidentally passing the wrong variable in `checkpointer`? I'm not even sure if we can test that (e.g., writing `u=state.v` instead of `u=state.u`). Probably it will be better if this is somehow automated and you just pass state and we look for the name match? And for the ones that have a `d` in the end, we extend the matching for allow that?

mcgibbon:
> Are we concerned about accidentally passing the wrong variable in `checkpointer`? I'm not even sure if we can test that (e.g., writing `u=state.v` instead of `u=state.u`). Probably it will be better if this is somehow automated and you just pass state and we look for the name match? And for the ones that have a `d` in the end, we extend the matching for allow that?

This isn't possible to fully automate, because the state names are frequently different from the saved names (I'd guess about 5% of the time) and we don't want to pin our variable names to the Fortran names. I'm actually more concerned about calling the checkpointer at a slightly wrong point in time compared to the Fortran. To a certain extent we will have to catch these by seeing regression failures, looking at the data, and being able to tell we've passed it the wrong variable or at the wrong time.

We might "automate" it on a case-by-case basis by constructiing a `kwargs` dictionary from e.g. `DycoreState` and passing it in to the checkpointer, in a helper method like `self._checkpoint_c_sw_in(dycore_state)`. Such helper methods can be made specific or general as needed.

fv3core/fv3core/stencils/dyn_core.py
```diff
@@ -525,6 +525,86 @@ def __init__(
             self.comm, grid_indexing, backend=stencil_factory.backend
         )
 
+    def _checkpoint_csw_in(self, state):
```
elynnwu, line 528: I see that you use a tag `in/out` for `_checkpoint_fvdynamics`. Is there a reason you cannot do them for `c_sw` and `d_sw`? It looks to me that the variables are the same.

mcgibbon: No reason. Updated them to use the same "tag" style.

elynnwu (approved):

mcgibbon:
launch jenkins

PR #191, opened 2022-05-20
Simpler setup

jdahm:
## Purpose

Users are confused by the repo. They need clearer and simpler instructions on how to use the code, even if it's still in a development phase.

## Code changes:

- Single entry-point for installing pace: `install.sh`
- README update

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,11 @@
 # Pace
 
-Pace is the top level directory that includes the FV3 dynamical core, physics, and util.
+> **WARNING** This repo is under active development and relies on code and data that is not publicly available at this point.
+
+Pace port of the FV3 model with physics. The project at the top-leve contains directories with each component of the model, including the dynamical core, physics, and util.
 
 If you are visiting for AMS 2022, we recommend you go to `driver/README.md`.
 
-**WARNING** This repo is under active development and relies on code and data that is not publicly available at this point.
 
 ## Getting started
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,10 +1,11 @@
 # Pace
 
-Pace is the top level directory that includes the FV3 dynamical core, physics, and util.
+> **WARNING** This repo is under active development and relies on code and data that is not publicly available at this point.
+
+Pace port of the FV3 model with physics. The project at the top-leve contains directories with each component of the model, including the dynamical core, physics, and util.
```
elynnwu, line 5:
```suggestion
Pace port of the FV3 model with physics. The project at the top-level contains directories with each component of the model, including the dynamical core, physics, util, and dsl tools shared among all components.
```

README.md
```diff
@@ -14,32 +15,32 @@ This git repository is laid out as a mono-repo, containing multiple independent
...
 ```
 
-To enter the container:
-```shell
-$ make dev
-```
+### Dynamical core tests
 
-Then in the container, dynamical core serial tests can be run:
+To run dynamical core tests, first get the test data from inside `fv3core` or `fv3gfs-physics` directory, then run the tests
```
elynnwu, line 31: Maybe say by default this will pull `c12_6ranks_standard`, which is the only one currently publicly available? Also, have we made the open buket?

docker/entrypoint.sh
```diff
@@ -1,11 +1,5 @@
...
 
 set -e
-pip install -r requirements.txt -c constraints.txt
-pip install -e /pace-util -c constraints.txt
-pip install -e /fv3core -c constraints.txt
-pip install -e /fv3gfs-physics -c constraints.txt
-pip install -e /stencils -c constraints.txt
-pip install -e /dsl -c constraints.txt
-pip install -e /driver -c constraints.txt
+./install.sh
```
elynnwu, line 4: 💯

elynnwu: oh wait, I think we cannot change this because those folders are mounted inside docker for when we run CI on gce.

jdahm:
We decided to hold off on this for now and first make the bucket non-private. The images used in pace need a larger refactor at some point anyway to avoid duplication.

PR #198, opened 2022-06-01
Create caches by initializing pace.driver.Driver

jdahm:
## Purpose

Initialize pace caches by initializing the Driver object.

## Post merge actions

Remove the Jenkins plans [pace-fv3core-cache-setup](https://jenkins.ginko.ch/job/pace-fv3core-cache-setup/) and [pace-physics-cache-setup](https://jenkins.ginko.ch/job/pace-physics-cache-setup/)

Refactor the Jenkins performance plan.


.jenkins/pace_cache_generation.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "USAGE: .jenkins/pace_cache_generation.sh backend experiment [target_uri]"
+fi
+
+backend=$1
+experiment=$2
+
+# If the target_uri starts with "daint:" then the Jenkins scripts are used
+# Use the target_uri as the last command line argument if passed, else the default daint path
+default_target_uri="daint:/scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$experiment/${backend//:/_}"
+target_uri=${1:-$daint_target_uri}
```
jdahm, line 15:
```suggestion
target_uri=${3:-$daint_target_uri}
```

.jenkins/pace_cache_generation.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+script_dir="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+pace_dir=$script_dir/../
+[[ "${NODE_NAME}" == *"daint"* ]] && source ~/.bashrc
+
+export LONG_EXECUTION=1
+
+if [[ $use_jenkins_action == "true" ]]; then
+    .jenkins/jenkins.sh initialize_driver $backend $experiment
```
jdahm, line 32:
```suggestion

if [[ $use_jenkins_action == "true" ]]; then
    export LONG_EXECUTION=1
    .jenkins/jenkins.sh initialize_driver $backend $experiment
```

jdahm, line 32: Need to investigate how Jenkins runner options are set.

.jenkins/pace_cache_generation.sh
```diff
@@ -0,0 +1,52 @@
...
+else
+    EXPERIMENT=$experiment TARGET=driver make get_test_data
+    $pace_dir/.jenkins/initialize_driver.py test_data/8.1.0/$experiment/driver $backend
```
jdahm, line 35: If `make get_test_data` outputs the version then we don't need to duplicate that logic here.

.jenkins/pace_cache_generation.sh
jdahm: I don't think we need any other options set.

.jenkins/actions/initialize_driver.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+backend=$1
+experiment=$2
+
+data_version=$(cd fv3core && EXPERIMENT=$experiment TARGET=driver make get_test_data | tail -1)
+.jenkins/initialize_driver.py test_data/$data_version/$experiment/driver $backend
```
jdahm, line 7:
```suggestion
.jenkins/initialize_driver.py fv3core/test_data/$data_version/$experiment/driver $backend
```

.jenkins/pace_cache_generation.sh
```diff
@@ -0,0 +1,50 @@
...
+if (( $# < 3 )); then
+    echo "USAGE: .jenkins/pace_cache_generation.sh backend experiment [target_uri]"
+fi
+
+backend=$1
+experiment=$2
+
+# If the target_uri starts with "daint:" then the Jenkins scripts are used
+# Use the target_uri as the last command line argument if passed, else the default daint path
+default_target_uri="daint:/scratch/snx3000/olifu/jenkins/scratch/store_gt_caches/$experiment/${backend//:/_}"
```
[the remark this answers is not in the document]
mcgibbon: I would suggest breaking out whether or not to use jenkins.sh to run the jenkins as a developer argument or environment variable, instead of folding it in to the path name. This way you can put a comment like "used to run this workflow locally bypassing jenkins.sh" and a reader can know they don't need to understand it in most cases.

As it is, I was pretty confused first reading about this path argument.

driver/pace/driver/__init__.py
```diff
@@ -2,6 +2,7 @@
     CreatesComm,
     CreatesCommSelector,
     MPICommConfig,
+    NullCommConfig,
```
mcgibbon, line 5: Thanks!

fv3core/Makefile
```diff
@@ -142,7 +142,8 @@ get_test_data:
 	rm -rf $(TEST_DATA_HOST) && \
 	$(MAKE) sync_test_data && \
 	$(MAKE) unpack_test_data ;\
-	fi
+	fi; \
+	echo $(FORTRAN_SERIALIZED_DATA_VERSION)
```
mcgibbon, line 146: You can leave this if you like, pointing it out in case this was meant to be temporary.

mcgibbon (commented):
Some minor suggestions for clarity, the biggest one is splitting `use_jenkins_action` into its own argument instead of folding it into a file path.

mcgibbon (approved):
Meant to approve, though please do review the suggested changes.

.jenkins/actions/initialize_driver.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
jdahm: Yeah good point. That was a quick and poor naming choice.

fv3core/Makefile
jdahm: This was actually intended. It allows scripts to extract the updated version string without duplicating logic.

mcgibbon: Since printing this version is an important side-effect, please add a comment saying this so it doesn't get deleted by someone who doesn't know.

jdahm:
@elynnwu determined that what this PR does so far is going to be too slow for our workflow. Options she came up with are:

1. Initialize the driver in parallel using 54 ranks. Pro: fast. Con: very wasteful.
2. Loop over the 9 combinations and submit jobs in parallel. Pro: fast and elegant. Con: Now we have to have batch logic in the script, which means reproducing the workflow locally without slurm (e.g. on local dev laptop) requires some changes.

In either case we need to agglomerate all 9 unique cache directories into one and symlink or copy to each rank's cache dir.

The optimal solution that we'd like to move towards is initializing a Driver for a single-tile run with 9 ranks, either in serial or parallel.

elynnwu:
> @elynnwu determined that what this PR does so far is going to be too slow for our workflow. Options she came up with are:
> 
> 1. Initialize the driver in parallel using 54 ranks. Pro: fast. Con: wasteful.
> 2. Loop over the 9 combinations and submit jobs in parallel. Pro: fast and elegant. Con: Now we have to have batch logic in the script, which means reproducing the workflow locally without slurm (e.g. on local dev laptop) requires some changes.
> 
> In either case we need to agglomerate all 9 unique cache directories into one and symlink or copy to each rank's cache dir.
> 
> The optimal solution that we'd like to move towards is initializing a Driver for a single-tile run with 9 ranks, either in serial or parallel.

For 2., we can call [launch job](https://github.com/ai2cm/buildenv/blob/df5b2b5a56885f9ddfd46a064cba1da77c76ea53/schedulerTools.sh#L32) directly, and it will take care of the logic to make sure the script is complete. But we still need to add the logic to check whether all 9 jobs succeeded.

mcgibbon:
Do we need to update the performance plan if we're removing standalone?

jdahm:
Yes, multiple Jenkins plans have to change before we can merge this. I'll add it to the To do. Thanks for pointing that out!

fv3core/Makefile
mcgibbon: Still need this.

jdahm: I added it above...

mcgibbon: I do worry about someone not noticing a comment 6 lines away when deleting this, if you can move it down to line 150-153. If not that's fine.

jdahm:
Closing for now since we can't get the caches to build in a reasonable time on piz daint.

PR #200, opened 2022-06-02
Add SnapshotCheckpointer

mcgibbon:
## Purpose

This PR adds a Checkpointer implementation which can be used to save datasets showing the evolution of variables between checkpointer calls.

## Code changes:

- Added pace.util.SnapshotCheckpointer

## Checklist

### DIFF
--- a/pace-util/pace/util/checkpointer/snapshots.py
+++ b/pace-util/pace/util/checkpointer/snapshots.py
@@ -1,7 +1,8 @@
 import collections
 
 import numpy as np
-import xarray as xr
+
+from pace.util._optional_imports import xarray as xr
 
 from .base import Checkpointer
 
@@ -28,7 +29,7 @@ class Snapshots:
         self._python_data[variable_name].append(python_data)
 
     @property
-    def dataset(self) -> xr.Dataset:
+    def dataset(self) -> "xr.Dataset":
         data_vars = {}
         for variable_name, savepoint_list in self._savepoints.items():
             savepoint_dim = f"sp_{variable_name}"
@@ -36,7 +37,12 @@ class Snapshots:
             data_vars[f"{variable_name}"] = make_dims(
                 savepoint_dim, variable_name, self._python_data[variable_name]
             )
-        return xr.Dataset(data_vars=data_vars)
+        if xr is None:
+            raise ModuleNotFoundError(
+                "xarray must be installed to use Snapshots.dataset"
+            )
+        else:
+            return xr.Dataset(data_vars=data_vars)
 
 
 class SnapshotCheckpointer(Checkpointer):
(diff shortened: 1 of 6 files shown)


pace-util/pace/util/checkpointer/snapshots.py
```diff
@@ -0,0 +1,72 @@
...
+    dims = [savepoint_dim] + [f"{label}_dim{i}" for i in range(len(data.shape[1:]))]
+    return dims, data
+
+
+class Snapshots:
+    def __init__(self):
+        self._savepoints = collections.defaultdict(list)
+        self._python_data = collections.defaultdict(list)
+
+    def compare(self, savepoint_name: str, variable_name: str, python_data):
```
oelbert, line 27: `compare` doesn't quite seem like the right name for this method to me. We're adding data to the snapshot, not looking at the differences between two datasets inside this method. `add_data` isn't as eloquent but something like that might be clearer

pace-util/pace/util/checkpointer/snapshots.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, rank: int):
+        if xr is None:
+            raise ModuleNotFoundError(
+                "xarray must be installed to use SnapshotCheckpointer"
+            )
+        self._rank = rank
+        self._snapshots = Snapshots()
+
+    def __call__(self, savepoint_name, **kwargs):
```
oelbert, line 62: Mildly curious why this is a __call__

oelbert (commented):
Looks good. One small nitpick and a question, otherwise good to go

pace-util/pace/util/checkpointer/snapshots.py
mcgibbon: I'll update it. This is a relic of when it was actually aligning Python and Fortran data (similarly to the _python_data name) and I wasn't careful because it's private API.

pace-util/pace/util/checkpointer/snapshots.py
mcgibbon: I'd like to keep this as-is for this PR since it's a larger change (changing the base class as well), I'm also not sure whether it should be changed or not.

oelbert: I don't think this needs to be changed, just good to know why.

pace-util/pace/util/checkpointer/snapshots.py
oelbert: Cool, thanks.

oelbert (approved):
Awesome

mcgibbon:
launch jenkins

mcgibbon:
launch jenkins

PR #220, opened 2022-06-21
Collect build and execution timings

jdahm:
## Purpose

Pace takes a while to compile, but so far we have done very little to diagnose. This PR will allow us to start to introspect both compilation times as well as execution times across Pace.

## Changes

Adds `show_build_report` and `show_exec_report` to `StencilFactory`. Internally the stencil objects hold a reference to a timing collector and add their `build_info` and `exec_info` to this.


driver/tests/mpi/run_save_and_load_restart.sh
```diff
@@ -3,9 +3,10 @@ set -e
 MPIRUN_CALL=${MPIRUN_CALL:-mpirun -n 6}
 cp examples/configs/baroclinic_c12_write_restart.yaml baroclinic_c12_write_restart.yaml
 cp examples/configs/baroclinic_c12_write_restart.yaml baroclinic_c12_run_two_steps.yaml
-sed -i 's/seconds: 225/seconds: 450/' baroclinic_c12_run_two_steps.yaml
-sed -i 's/save_restart: true/save_restart: false/' baroclinic_c12_run_two_steps.yaml
-sed -i 's/path: "output.zarr"/path: "run_two_steps_output.zarr"/' baroclinic_c12_run_two_steps.yaml
+sed -i.bak 's/seconds: 225/seconds: 450/' baroclinic_c12_run_two_steps.yaml
+sed -i.bak 's/save_restart: true/save_restart: false/' baroclinic_c12_run_two_steps.yaml
+sed -i.bak 's/path: "output.zarr"/path: "run_two_steps_output.zarr"/' baroclinic_c12_run_two_steps.yaml
+rm *.bak
```
jdahm, line 9: What was here didn't work on BSD sed, so I changed it to be compatible in both cases.

driver/tests/mpi/test_restart.py
```diff
@@ -81,4 +81,4 @@ def test_restart():
                     restart_physics[var].values,
                 )
     finally:
-        shutil.rmtree("RESTART")
+        shutil.rmtree("RESTART", ignore_errors=True)
```
jdahm, line 84: try-finally should have an `ignore_errors=True`. Again, unrelated to this PR but a tiny change we need to push though.

dsl/pace/dsl/stencil.py
```diff
@@ -334,9 +388,18 @@ def __init__(
             self.stencil_object = self._compile()
 
     def _compile(self):
+
+        # Ensure the stencil names are unique
```
jdahm, line 392:
```suggestion
```

FlorianDeconinck (approved):
I like it, but can we have it auto-print the report on a configuration variable? Maybe on the driver or even per stencil (for translate test), so we can turn it on more easily

dsl/pace/dsl/stencil.py
```diff
@@ -202,6 +202,52 @@ def report_diff(arg: np.ndarray, numpy_arg: np.ndarray, label) -> str:
     return report
 
 
+@dataclasses.dataclass
+class TimingCollector:
+    build_info: Dict[str, dict] = dataclasses.field(default_factory=dict)
+    exec_info: Dict[str, dict] = dataclasses.field(default_factory=dict)
```
mcgibbon, line 208: Please add a docstring for the class, with an `Attributes:` section that describes what the keys and values of these dictionaries are. You can use something brief for the values (e.g. "a gt4py build_info dictionary") but it would help a lot to know if these keys are stencil names, keys of build_info, or something completely different.

dsl/pace/dsl/stencil.py
```diff
@@ -290,13 +342,15 @@ def __init__(
...
             skip_passes: compiler passes to skip when building stencil
+            timing_collector: Optional object that accumulates timings
         """
         if isinstance(origin, tuple):
             origin = cast_to_index3d(origin)
         origin = cast(Union[Index3D, Mapping[str, Tuple[int, ...]]], origin)
         self.origin = origin
         self.domain: Index3D = cast_to_index3d(domain)
         self.stencil_config: StencilConfig = stencil_config
+        self.timing_collector = timing_collector or TimingCollector()
```
mcgibbon, line 353: It's better to avoid relying on the implementation of `__nonzero__` for timing_collector by explicitly checking if it is None. Can you also make this a private attribute?

dsl/pace/dsl/stencil.py
```diff
@@ -263,6 +310,10 @@ def __call__(
         )
 
 
+def _stencil_object_name(stencil_object: gt4py.StencilObject) -> str:
+    return type(stencil_object).__name__
```
mcgibbon, line 314: Is this going to give unique names for each stencil? If it is, can you add a `#` comment to that effect, e.g. saying the dynamically-generated gt4py stencil class names include a hash of the compile-time constants?

dsl/pace/dsl/stencil.py
```diff
@@ -365,6 +426,10 @@ def __call__(self, *args, **kwargs) -> None:
         if self.stencil_object is None:
             self.stencil_object = self._compile()
 
+        exec_info = self.timing_collector.exec_info.setdefault(
+            _stencil_object_name(self.stencil_object), {}
+        )
```
mcgibbon, line 431: You can avoid this setdefault if you initialize exec_info as a `collections.defaultdict(dict)`. You can only do this if you make exec_info private (i.e. not possible to pass as an init argument of TimingCollector) to guarantee it's a defaultdict and not some other dict.

dsl/pace/dsl/stencil.py
```diff
@@ -871,6 +940,7 @@ def __init__(self, config: StencilConfig, grid_indexing: GridIndexing):
         """
         self.config: StencilConfig = config
         self.grid_indexing: GridIndexing = grid_indexing
+        self.timing_collector = TimingCollector()
```
mcgibbon, line 943: Can this attribute be private?

dsl/pace/dsl/stencil.py
```diff
@@ -895,15 +965,18 @@ def from_origin_domain(
         """
         if self.config.compare_to_numpy:
             cls: Type = CompareToNumpyStencil
+            extra_kwargs = {}
         else:
             cls = FrozenStencil
+            extra_kwargs = {"timing_collector": self.timing_collector}
```
mcgibbon, line 971: Why do we only collect timings if not comparing to numpy? These seem like orthogonal configuration options.

dsl/pace/dsl/stencil.py
```diff
@@ -953,14 +1026,22 @@ def restrict_vertical(self, k_start=0, nk=None) -> "StencilFactory":
             grid_indexing=self.grid_indexing.restrict_vertical(k_start=k_start, nk=nk),
         )
 
+    def show_build_report(self, key: str = "build_time", **kwargs) -> str:
+        """Report all stencils built by this factory."""
+        return self.timing_collector.show_build_report(key, **kwargs)
+
+    def show_exec_report(self, key: str = "total_run_time", **kwargs) -> str:
+        """Report all stencils executed that were built by this factory."""
+        return self.timing_collector.show_exec_report(key, **kwargs)
```
mcgibbon, line 1035: Please add tests of these functions. I think the code has a bug (key here is a descriptor, but in the lower levels it's a stencil class name), but it's easier to say if it does or doesn't if there's a test.

jdahm: Yes, I'll add a test. Looks correct to me:
```python
class TimingCollector:
...
    def show_exec_report(self, key: str = "total_run_time", **kwargs) -> str:
        return type(self)._show_report(self.exec_info, key, **kwargs)
```

dsl/pace/dsl/stencil.py
jdahm: Hmm, I tried that but I think I prefer this solution because it allows `exec_info` to remain public.

mcgibbon: you can still make it readable-public with a property method that returns `self._exec_info`, it's just that you can't have it as an init argument. I don't see any use case for passing it as an init argument, anyways.

.circleci/config.yml
```diff
@@ -59,6 +59,40 @@ jobs:
...
+          keys:
+            - v2-{{ checksum "constraints.txt" }}-{{ checksum "dsl/setup.py" }}
+      - run:
+          name: run tests
+          command: |
+            if [ $(python changed_from_main.py dsl) != false ]; then
+              pip3 install --upgrade setuptools wheel
+              python3 -m venv dsl/venv
+              . dsl/venv/bin/activate
+              pip3 install -r dsl/requirements.txt -c constraints.txt
```
mcgibbon, line 81: Where is `dsl` installed? I don't see a command for it.

dsl/pace/dsl/stencil.py
```diff
@@ -953,14 +1043,22 @@ def restrict_vertical(self, k_start=0, nk=None) -> "StencilFactory":
             grid_indexing=self.grid_indexing.restrict_vertical(k_start=k_start, nk=nk),
         )
 
+    def show_build_report(self, key: str = "build_time", **kwargs) -> str:
```
mcgibbon, line 1046: nit: you could remove "show" from these method names (possibly replacing with "get"), since this doesn't actually show (print) a report, just returns it as a string.

.circleci/config.yml
jdahm: `dsl` is just the dsl directory in pace.

jdahm: Added it!

PR #222, opened 2022-06-22
Automatically calculate translate thresholds using perturbations

mcgibbon:
## Purpose

This PR introduces code to output suggested translate test error thresholds for each variable when a test fails. This is then used to update thresholds for failures that occur on baremetal M1 mac.

The QSInit thresholds are updated based solely on the error being small, since it actually has no inputs and perturbations cannot be used to suggest a threshold for it.

This code could later be used to automatically generate a threshold override file, or to use stricter thresholds for different variables, but this work is out of scope of this PR.

## Infrastructure changes:

- Sequential savepoint tests now output suggested translate test error thresholds for each variable when a test fails.
- Updated error thresholds for Riem_Solver3, QSInit, and NH_P_Grad

## Checklist


stencils/pace/stencils/testing/test_translate.py
```diff
@@ -167,6 +168,46 @@ def process_override(threshold_overrides, testobj, test_name, backend):
             )
 
 
+def perturb(input):
+    roundoff = 1e-16
+    for data in input.values():
+        if isinstance(data, np.ndarray) and data.dtype in (np.float64, np.float32):
+            not_fill_value = data < 1e10
```
elynnwu, line 175: Is this safe? I'm thinking we don't have anything larger than 1e10, but isn't fill value usually ~1e30?

stencils/pace/stencils/testing/test_translate.py
```diff
@@ -167,6 +168,46 @@ def process_override(threshold_overrides, testobj, test_name, backend):
...
+            # multiply data by roundoff-level error
+            data[not_fill_value] *= 1.0 + np.random.uniform(
+                low=-roundoff, high=roundoff, size=data[not_fill_value].shape
+            )
+
+
+N_THRESHOLD_SAMPLES = 10
```
elynnwu, line 182: Did you try a couple of numbers and the threshold stopped ~changing after a few?

stencils/pace/stencils/testing/test_translate.py
mcgibbon: I could use a larger threshold, I should do a (automatic, in code here) search to see first whether any other fill values are used. I thought there were values other than 1e30, sincet the reference Fortran also sometimes uses some random fill values in its code.

stencils/pace/stencils/testing/test_translate.py
mcgibbon: I just put 10 and called it good. There's a tradeoff to make between statistical robustness and performance. I'll experiment with this a bit and see how it changes, but I doubt I can use much smaller than this, and we don't need to use a larger one if we use the right scale factor (we probably should set threshold a bit larger than the greatest difference we observe). In these cases I set the threshold about 2-5x larger than the "suggested" (so 5e-10 instead of 1e-10, for example).

mcgibbon: I tested some more values on Riem_Solver3, but I think I'll just leave it at 10. Even at 20 runs, it wasn't enough to really nail down a max relative difference for `w`, but 10 was enough to have consistent output differences for every other value.

elynnwu (approved):

mcgibbon:
launch jenkins

PR #252, opened 2022-07-25
Add "pair_debug" configuration option for driver

mcgibbon:
## Purpose

One model error we can run into is race conditions. Race conditions result in non-deterministic model output. This PR adds a "pair_debug" execution mode to the driver where it will run two copies of the model simultaneously, constantly comparing the inputs and outputs of stencils between these two copies and raising an exception when a difference occurs, indicating non-deterministic output.

## Code changes:

- Added "pair_debug" boolean flag to DriverConfig which enables or disables duplicate-model debugging (disabled by default). This option may not be used simultaneously with read/write MPI execution due to unimplemented features, but could be supported by a later PR
- Added `sendrecv` method to Comm abstract base class, required for pair model comparison

## Checklist


dsl/pace/dsl/stencil.py
```diff
@@ -934,7 +979,7 @@ def get_quantity_halo_spec(
 class StencilFactory:
     """Configurable class which creates stencil objects."""
 
-    def __init__(self, config: StencilConfig, grid_indexing: GridIndexing):
+    def __init__(self, config: StencilConfig, grid_indexing: GridIndexing, comm):
```
elynnwu, line 982: Is it weird that `comm` is a required argument to `StencilFactory`? I think it kind of has to otherwise if would be strange to default it to None which will make it more confusing.

dsl/pace/dsl/stencil.py
```diff
@@ -351,6 +377,7 @@ def __init__(
         externals: Optional[Mapping[str, Any]] = None,
         skip_passes: Tuple[str, ...] = (),
         timing_collector: Optional[TimingCollector] = None,
+        comm=None,
```
elynnwu, line 380: I notice that you default to None here but not for `StencilFactory`. Is it because we don't call `FrozenStencil` directly so we don't expect the users to ever have to see this?

dsl/pace/dsl/stencil.py
mcgibbon: No, it should be optional. Updating.

dsl/pace/dsl/stencil.py
mcgibbon: Updated StencilFactory.

elynnwu (approved):
Do we want to add a note that you should specify 2x the ranks when pair_debug is turned on?

mcgibbon:
launch jenkins

mcgibbon:
launch jenkins

PR #267, opened 2022-08-09
Add LICENSE.md

ofuhrer:
## Purpose

Add a software license to repository which reflects the license of the parent project (FV3GFS from NCEP).


twicki (approved):
I believe this should work. Might need a double-check from legal?

PR #268, opened 2022-08-10
Extract gt4py setup from compilation config init

twicki:
## Purpose

The initialization of a `CompilationConfig` object should not have side-effects. These are removed here.

## Code changes:

- moved `decomposition.py` and it's test from `pace/utils` to `pace/dsl`
- created `configure_gt4py` function for StencilConfig
  - added calls to it where appropriate

## Checklist

### DIFF
--- a/driver/pace/driver/driver.py
+++ b/driver/pace/driver/driver.py
@@ -348,8 +348,8 @@ class Driver:
             device_sync=original_config.device_sync,
             run_mode=original_config.run_mode,
             use_minimal_caching=original_config.use_minimal_caching,
-            communicator=communicator,
         )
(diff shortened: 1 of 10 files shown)


mcgibbon:
To summarize our discussion, there are two attributes on StencilConfig which require a communicator to determine. In order to initialize StencilConfig without a communicator we will need to separate those two attributes and pass them separately to StencilFactory. This is a much larger change than I want to ask you to make, and there might still be gotchas I'm missing. I will try to do this refactor later with your branch as a starting point.

For safety, I may also add a check that the global gt4py config matches the passed config when StencilFactory is initialized, to avoid the possibility that we forget to call the gt4py configuration method.

PR #273, opened 2022-08-16
Ensure kslice_from_inputs makes a slice of ints

jdahm:
## Purpose

The `MetricTerms` class agrid init fails on 54 ranks.

## Code changes:

- Cast nk to an integer so slices don't have floats in `kslice_from_inputs`.


mcgibbon (approved):
LGTM, thanks!

PR #275, opened 2022-08-16
Run init tests on savepoint_tests_mpi

jdahm:
## Purpose

@oelbert and I noticed that the Grid tests are not run in CI 😳 😨 😱 😠 , since the CI only uses the default `TARGET` and the grid init was put into `TARGET=init` instead of `TARGET=dycore`.

## Code changes:

- Add tests.
- Cast `nk` to an `int` to fix one grid-related test.

### DIFF
--- a/fv3core/.jenkins/actions/run_parallel_regression_tests.sh
+++ b/fv3core/.jenkins/actions/run_parallel_regression_tests.sh
@@ -9,12 +9,16 @@ make get_test_data
 
 if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="" MPIRUN_ARGS="" make savepoint_tests_mpi
+    TARGET=init CONTAINER_CMD="" MPIRUN_ARGS="" make savepoint_tests_mpi
 else
     make savepoint_tests_mpi
+    TARGET=init make savepoint_tests_mpi
 fi
 export TEST_ARGS="${TEST_ARGS} --compute_grid"
 if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="" MPIRUN_ARGS="" make savepoint_tests_mpi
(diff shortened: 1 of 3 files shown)


fv3core/.jenkins/actions/run_parallel_regression_tests.sh
```diff
@@ -9,12 +9,16 @@ make get_test_data
 
 if [ ${python_env} == "virtualenv" ]; then
     CONTAINER_CMD="" MPIRUN_ARGS="" make savepoint_tests_mpi
+    TARGET=init CONTAINER_CMD="" MPIRUN_ARGS="" make savepoint_tests_mpi
```
jdahm, line 12: We probably don't `get_test_data` with `TARGET=init`, so we have to make sure that is done before merging this.

jdahm:
... and the translate tests of course fail because of https://github.com/ai2cm/pace/issues/276.

fv3core/.jenkins/actions/run_parallel_regression_tests.sh
elynnwu: On daint you have to either download ahead of time (it is already downloaded) or run jenkins twice (which usually is the case) because each config would try to download and try and write to the same directory.

jdahm: Does that mean we need to make more changes to this PR?

elynnwu: Should be good to go now.

elynnwu (approved):

jdahm:
It doesn't look like this adds run-time to the fv3core tests in any appreciable manner, since the tests still take ~45 minutes.

mcgibbon (approved):

PR #283, opened 2022-08-17
Fix the model inconsistencies between 6 and 54 rank runs

twicki:
## Purpose

We have seen inconsistencies between 6 and 54 rank runs. These should not eixist and get solved by this.

## Code changes:

- Add the `allreduce` method to Comm, MPIComm and NullComm
- Assert LocalComm in `allreduce` because it requires synchronicity. 
- Removed the try-except block in `generation.py` in the `_reduce_global_area_minmaxes` function.
- updated `gt4py_utils.py` to properly handle scalars

## Checklist

### DIFF
--- a/dsl/pace/dsl/gt4py_utils.py
+++ b/dsl/pace/dsl/gt4py_utils.py
@@ -355,9 +355,13 @@ def asarray(array, to_type=np.ndarray, dtype=None, order=None):
             return np.asarray(array, dtype, order)
         else:
             return cp.asarray(array, dtype, order)
-    if cp and hasattr(array, "data") and (
-        isinstance(array, memoryview)
-        or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
+    if (
+        cp
+        and hasattr(array, "data")
+        and (
+            isinstance(array, memoryview)
+            or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
+        )
     ):
         if to_type is np.ndarray:
             order = "F" if order is None else order
--- a/dsl/pace/dsl/gt4py_utils.py
+++ b/dsl/pace/dsl/gt4py_utils.py
@@ -355,7 +355,7 @@ def asarray(array, to_type=np.ndarray, dtype=None, order=None):
             return np.asarray(array, dtype, order)
         else:
             return cp.asarray(array, dtype, order)
-    if cp and (
+    if cp and hasattr(array, "data") and (
         isinstance(array, memoryview)
         or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
     ):
--- a/dsl/pace/dsl/gt4py_utils.py
+++ b/dsl/pace/dsl/gt4py_utils.py
@@ -355,12 +355,11 @@ def asarray(array, to_type=np.ndarray, dtype=None, order=None):
             return np.asarray(array, dtype, order)
         else:
             return cp.asarray(array, dtype, order)
-    if (
-        cp
-        and hasattr(array, "data")
-        and (
-            isinstance(array, memoryview)
-            or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
+    if cp and (
+        isinstance(array, memoryview)
+        or (
+            hasattr(array, "data")
+            and isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
         )
     ):
         if to_type is np.ndarray:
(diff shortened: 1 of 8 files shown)


pace-util/pace/util/caching_comm.py
```diff
@@ -49,12 +49,14 @@ class CachingCommData:
...
     bcast_objects: List[Any] = dataclasses.field(default_factory=list)
     received_buffers: List[np.ndarray] = dataclasses.field(default_factory=list)
+    generic_obj_buffers: List[Any] = dataclasses.field(default_factory=list)
     split_data: List["CachingCommData"] = dataclasses.field(default_factory=list)
 
     def __post_init__(self):
         self._i_bcast = 0
         self._i_buffers = 0
         self._i_split = 0
+        self._i_generic_obj = 0
```
mcgibbon, line 59: "generic" has a specific meaning, I'd suggest `python_obj` or even just `obj` (all python classes inherit from `object` so this is accurate).

pace-util/pace/util/comm.py
```diff
@@ -63,3 +63,7 @@ def Irecv(self, recvbuf, source, tag: int = 0, **kwargs) -> Request:
     @abc.abstractmethod
     def Split(self, color, key) -> "Comm":
         ...
+
+    @abc.abstractmethod
+    def allreduce(self, sendobj, op=None) -> Any:
```
mcgibbon, line 68: `None` is not a valid argument for `op`, unfortunately. The default value is `MPI.SUM`, which is only available if we can import from mpi4py, but I think we can get away with the built-in `sum` object as default.
```suggestion
    def allreduce(self, sendobj, op=sum) -> Any:
```

I'd also be fine with making this a required argument on our end.

dsl/pace/dsl/gt4py_utils.py
```diff
@@ -355,9 +355,13 @@ def asarray(array, to_type=np.ndarray, dtype=None, order=None):
...
-    if cp and (
-        isinstance(array, memoryview)
-        or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
+    if (
+        cp
+        and hasattr(array, "data")
+        and (
+            isinstance(array, memoryview)
+            or isinstance(array.data, (cp.ndarray, cp.cuda.memory.MemoryPointer))
+        )
```
mcgibbon, line 364: The logic for this has changed, can you make sure it still evaluates to `True` if the object is a `memoryview`, even though memoryview objects have no "data" attribute?

You may be able to make use of `getattr(array, "data", None)` which will default to None if there is no attribute.

mcgibbon (changes requested):

pace-util/pace/util/comm.py
twicki: none is a valid argument for op

pace-util/pace/util/caching_comm.py
twicki: what is this specific meaning that we are not upholding with the logic in this class?

dsl/pace/dsl/gt4py_utils.py
twicki: fixed

pace-util/pace/util/caching_comm.py
mcgibbon: The types being stored are not specifically "generic types" (e.g. as discussed [here](https://docs.oracle.com/javase/tutorial/java/generics/types.html)).

pace-util/pace/util/null_comm.py
```diff
@@ -86,3 +86,6 @@ def Split(self, color, key):
             comm.total_ranks = total_ranks
         self._split_comms[color].append(new_comm)
         return new_comm
+
+    def allreduce(self, sendobj, op=None) -> Any:
+        return self._fill_value
```
mcgibbon, line 91:
```suggestion
    def allreduce(self, sendobj, op=None) -> Any:
        return sendobj
```
nit: data doesn't matter for null comms, but this would be less likely to cause an exception.

pace-util/HISTORY.md
```diff
@@ -9,9 +9,11 @@ Major changes:
 - Added Checkpointer and NullCheckpointer classes
 - Added SnapshotCheckpointer
 - Comm and Request abstract base classes are added to the top level
+- Added the following attributes/methods to the Comm abstract base classes: `allreduce`
 
 Minor changes:
 - Deleted deprecated `finish_halo_update` method from CubedSphereCommunicator
+- fixed a bug in `pace.util.grid.` where `_reduce_global_area_minmaxes` would use local values instead of the gathered ones
```
mcgibbon, line 16:
```suggestion
- fixed a bug in `pace.util.grid` where `_reduce_global_area_minmaxes` would use local values instead of the gathered ones
```

mcgibbon (approved):
Thanks for the fix!

mcgibbon:
When you merge can you update the PR title to describe what was fixed, so the merge commit says what was changed?

twicki:
good point, done

PR #290, opened 2022-08-22
do not build physics if disabled

twicki:
## Purpose

We are currently still building the entire physics package even if we're not running it. This solves that problem

## Code changes:

- Change `driver.py` to not build physics components if they are not used


elynnwu (approved):
lgtm! will merge this in the other branch

PR #316, opened 2022-09-08
Feature/transform grid stretch

ajdas1:
## Purpose

Added function that allows for Schmidt grid transformation - provided a stretch factor and target longitude and latitude.
Tile 6 (S pole) is shrunk by stretch_factor (all other tiles are stretched to compensate), and then tile 6 is shifted so it is centered on target longitude and latitude.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,3 @@ RESTART/
 _dacegraphs
 
 **/testing/output/*
-
-*nc
-*png
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,6 @@ RESTART/
 _dacegraphs
 
 **/testing/output/*
+
+*nc
+*png
--- a/driver/pace/driver/gridconfig.py
+++ b/driver/pace/driver/gridconfig.py
@@ -7,9 +7,9 @@ import numpy as np
 @dataclasses.dataclass
 class GridConfig:
     stretch_mode: bool = False
-    stretch_factor: Optional[np.float_] = 1.0
-    lon_target: Optional[np.float_] = 0.0
-    lat_target: Optional[np.float_] = 0.0
+    stretch_factor: Optional[np.float_] = None
+    lon_target: Optional[np.float_] = None
+    lat_target: Optional[np.float_] = None
 
     def __post_init__(self):
         if self.stretch_mode:
--- a/driver/pace/driver/gridconfig.py
+++ b/driver/pace/driver/gridconfig.py
@@ -1,12 +1,8 @@
 import dataclasses
-import os.path
-import subprocess
-from typing import List, Optional
+from typing import Optional
 
 import numpy as np
 
-from .report import collect_data_and_write_to_file
-from pace.util import Namelist
 
 @dataclasses.dataclass
 class GridConfig:
@@ -18,7 +14,9 @@ class GridConfig:
     def __post_init__(self):
         if self.stretch_mode:
             if not self.stretch_factor:
-                raise ValueError("Stretch_mode is true, but no stretch_factor is provided.")
+                raise ValueError(
+                    "Stretch_mode is true, but no stretch_factor is provided."
+                )
             if not self.lon_target:
                 raise ValueError("Stretch_mode is true, but no lon_target is provided.")
             if not self.lat_target:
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -162,3 +162,6 @@ RESTART/
 _dacegraphs
 
 **/testing/output/*
+
+*nc
+*png
```
elynnwu, line 167: I think the `.nc` files are probably fine to ignore, but we may in the future (if not already) have `.png` files as part of the repository. So I would not ignore the png files.

util/pace/util/grid/stretch_transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    - lat in radians
+    - stretch_factor (e.g. 3.0 means that the resolution on
+        tile 6 becomes 3 times as fine)
+    - lon_target in degrees (from namelist)
+    - lat_target in degrees (from namelist)
+
+    Outputs:
+    - lon_transform in radians
+    - lat_transform in radians
+    """
```
elynnwu, line 35: Can you format the doc string so that it matches the other places in pace?
Should look something like this:
```
        Description of function
        Args:
            x (inout): variable x
            y (in): variable y
        Returns:
            z (out): return variable z    
```

util/pace/util/grid/stretch_transformation.py
```diff
@@ -0,0 +1,129 @@
...
+from pace.util import Quantity
+
+
+def direct_transform(
+    *,
+    lon: Quantity,
+    lat: Quantity,
+    stretch_factor: float,
+    lon_target: float,
+    lat_target: float,
```
elynnwu, line 15: Should the type be `np.float_` or `pace.dsl.typing.Float`?

util/pace/util/grid/stretch_transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    - lon_transform in radians
+    - lat_transform in radians
+    """
+
+    if isinstance(lon, Quantity):
+        lon_data = lon.data
+        lat_data = lat.data
+    elif isinstance(lon, np.ndarray):
+        lon_data = lon
+        lat_data = lat
```
elynnwu, line 42: Based on the function signature, lat/lon is of type `Quantity`. Change the type hint to `Union[Quantity, np.ndarray]`?

elynnwu: Also you probably want to throw an exception if it's neither type?

util/pace/util/grid/stretch_transformation.py
```diff
@@ -0,0 +1,129 @@
+import copy as cp
```
elynnwu, line 1: We use `cp` strictly for `import cupy as np`, I suggest you just `import copy` and use `copy.X` everywhere.

util/pace/util/grid/stretch_transformation.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Use:
+    output = sign(input, pn)
+
+    Takes the sign of pn (positive or negative) and assigns it to value.
+
+    Inputs:
+    - input: value to be assigned a sign
+    - pn: value whose sign is assigned to input
+    """
```
elynnwu, line 108: Update doc string format here too

elynnwu (commented):
Also, you want to expose stretch factor, lat/lon_target to the namelist right? So that you can specify them from driver namelist.

driver/pace/driver/gridconfig.py
```diff
@@ -0,0 +1,23 @@
...
+    def __post_init__(self):
+        if self.stretch_mode:
+            if not self.stretch_factor:
+                raise ValueError(
+                    "Stretch_mode is true, but no stretch_factor is provided."
+                )
+            if not self.lon_target:
+                raise ValueError("Stretch_mode is true, but no lon_target is provided.")
+            if not self.lat_target:
+                raise ValueError("Stretch_mode is true, but no lat_target is provided.")
```
elynnwu, line 23: I think these won't error because you have default values. It should work if you just take out the default values? Not sure how dacite handles optional inputs, you will want to test this.

elynnwu (commented):
I don't think you are actually running the transformation, should be added in driver.

elynnwu (approved):
lgtm!

PR #319, opened 2022-09-09
DaCe Orchestration: fix parsing, de-inline grid data scalars & memory leak

FlorianDeconinck:
## Purpose

This PR address two issues:
* `main` has broken orchestration due to the passing of `wsd` / `cappa` as quantities down the stack. A DaCe bug means that `wsd` has output only creates a memory description failure we had to get around.
* Two damping coefficients scalars `da_min` and `da_min_c` are decomposition dependent, which breaks our distributed compile system with DaCe.
* Using internal of C_SW instead of return values to workaround a memory leak happening in the pooling of DaCe.

## Code changes:
- Turn distributed compilation back on, e.g. `DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False`
- `wsd` and `cappa` are now cached on the proper top-level module (fv_dyn & acoustics) and there storage are access via `self`
- See comment left in code:
```
# We need to use a getter for da_min & da_min_c in order to go around a DaCe inline
# behavior. As part of the automatic optimization process, DaCe tries to inline
# as many scalars as possible.
# The grid is _not_ passed as an input to the top level function we orchestrate,
# so its scalar values will be inlined.

# 'alas, our distributed compilation system works by compiling a 3,3 layout
# top tile, then using those 9 caches on every layout upward.
# This setup leads to the values of da_min/da_min_c from the 3,3 layout
# to be inlined in the generated code. Those variables are used in runtime
# calculation (kinetic energy, etc.) which obviously leads to misbehaving numerics
# and errors when the 3,3 layout values are used on larger layouts

# The solution we implement here is making use of the fact that callbacks
# are never inlined in dace optimization. the current workaround uses the
# following functions.

# An alternative would be to pass the Grid or the DampingCoefficients to DaCe,
# clearly flagging it has a dynamic piece of memory (which would
# cancel any inlining) but the feature to do that (dace.struct)
# is currently in disarray.
# N.B.: another solution is to pass da_min and da_min_c as input, put it seems
# odd and adds a lot of boilerplate throughout the model code.
```
- `dace_constants_args` has been renamed to `dace_compiletime_args` to match DaCe API naming change
- change in field referenced in dyn_core, see comment left in code above `self.ptc` and `self.delpc`:
```
# ToDo: Due to DaCe VRAM pooling creating a memory
# leak with the usage pattern of those two fields (to be fixed soon)
# We use the C_SW internal to workaround it e.g.:
#  - self.cgrid_shallow_water_lagrangian_dynamics.delpc
#  - self.cgrid_shallow_water_lagrangian_dynamics.ptc
```

## Requirements changes:

- N/A

## Infrastructure changes:

- N/A

## Checklist

- [ ] You have followed the coding standards guidelines established at [Code Review Checklist](https://drive.google.com/file/d/1R0nqOxfYnzaSdoYdt8yjx5J482ETI2Ft/view?usp=sharing).

### DIFF
--- a/dsl/pace/dsl/dace/build.py
+++ b/dsl/pace/dsl/dace/build.py
@@ -170,7 +170,9 @@ def get_sdfg_path(
             )
         # Check resolution per tile
         build_resolution = ast.literal_eval(build_info_file.readline())
-        if config.tile_resolution != build_resolution:
+        if (config.tile_resolution[0] / config.layout[0]) != (
+            build_resolution[0] / build_layout[0]
+        ):
             raise RuntimeError(
                 f"SDFG build for resolution {build_resolution}, "
                 f"cannot be run with current resolution {config.tile_resolution}"
--- a/dsl/pace/dsl/dace/dace_config.py
+++ b/dsl/pace/dsl/dace/dace_config.py
@@ -10,7 +10,7 @@ from pace.util.communicator import CubedSphereCommunicator
 # TODO (floriand): Temporary deactivate the distributed compiled
 # until we deal with the Grid data inlining during orchestration
 # See github issue #301
-DEACTIVATE_DISTRIBUTED_DACE_COMPILE = True
+DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
 
 
 class DaCeOrchestration(enum.Enum):
--- a/dsl/pace/dsl/dace/dace_config.py
+++ b/dsl/pace/dsl/dace/dace_config.py
@@ -7,9 +7,9 @@ from pace.dsl.gt4py_utils import is_gpu_backend
 from pace.util.communicator import CubedSphereCommunicator
 
 
-# TODO (floriand): Temporary deactivate the distributed compiled
-# until we deal with the Grid data inlining during orchestration
-# See github issue #301
+# This can be turned on to revert compilation for orchestration
+# in a rank-compile-itself more, instead of the distributed top-tile
+# mechanism.
 DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
 
 
--- a/dsl/pace/dsl/dace/dace_config.py
+++ b/dsl/pace/dsl/dace/dace_config.py
@@ -10,7 +10,7 @@ from pace.util.communicator import CubedSphereCommunicator
 # TODO (floriand): Temporary deactivate the distributed compiled
 # until we deal with the Grid data inlining during orchestration
 # See github issue #301
-DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
+DEACTIVATE_DISTRIBUTED_DACE_COMPILE = True
 
 
 class DaCeOrchestration(enum.Enum):
--- a/dsl/pace/dsl/dace/orchestration.py
+++ b/dsl/pace/dsl/dace/orchestration.py
@@ -435,8 +435,8 @@ def orchestrate(
         obj: object which methods is to be orchestrated
         config: DaceConfig carrying model configuration
         method_to_orchestrate: string representing the name of the method
-        dace_compiletime_args: list of names of arguments to be flagged has dace.constant
-                            for orchestration to behave
+        dace_compiletime_args: list of names of arguments to be flagged has
+                               dace.compiletime for orchestration to behave
     """
 
     if config.is_dace_orchestrated():
@@ -517,8 +517,8 @@ def orchestrate_function(
 
     Args:
         config: DaceConfig carrying model configuration
-        dace_compiletime_args: list of names of arguments to be flagged has dace.constant
-                            for orchestration to behave
+        dace_compiletime_args: list of names of arguments to be flagged has
+                               dace.compiletime for orchestration to behave
     """
 
     def _decorator(func: Callable[..., Any]):
--- a/dsl/pace/dsl/dace/orchestration.py
+++ b/dsl/pace/dsl/dace/orchestration.py
@@ -425,7 +425,7 @@ def orchestrate(
     obj: object,
     config: DaceConfig,
     method_to_orchestrate: str = "__call__",
-    dace_constant_args: List[str] = [],
+    dace_compiletime_args: List[str] = [],
 ):
     """
     Orchestrate a method of an object with DaCe.
@@ -435,7 +435,7 @@ def orchestrate(
         obj: object which methods is to be orchestrated
         config: DaceConfig carrying model configuration
         method_to_orchestrate: string representing the name of the method
-        dace_constant_args: list of names of arguments to be flagged has dace.constant
+        dace_compiletime_args: list of names of arguments to be flagged has dace.constant
                             for orchestration to behave
     """
 
@@ -444,7 +444,7 @@ def orchestrate(
             func = type.__getattribute__(type(obj), method_to_orchestrate)
 
             # Flag argument as dace.constant
-            for argument in dace_constant_args:
+            for argument in dace_compiletime_args:
                 func.__annotations__[argument] = DaceCompiletime
 
             # Build DaCe orchestrated wrapper
@@ -509,7 +509,7 @@ def orchestrate(
 
 def orchestrate_function(
     config: DaceConfig = None,
-    dace_constant_args: List[str] = [],
+    dace_compiletime_args: List[str] = [],
 ) -> Union[Callable[..., Any], _LazyComputepathFunction]:
     """
     Decorator orchestrating a method of an object with DaCe.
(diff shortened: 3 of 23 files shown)


dsl/pace/dsl/dace/dace_config.py
```diff
@@ -10,7 +10,7 @@
 # TODO (floriand): Temporary deactivate the distributed compiled
 # until we deal with the Grid data inlining during orchestration
 # See github issue #301
-DEACTIVATE_DISTRIBUTED_DACE_COMPILE = True
+DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
```
FlorianDeconinck, line 13: Comment on top of this will be changed before merge - doing a few tests first

FlorianDeconinck: Done

dsl/pace/dsl/dace/build.py
```diff
@@ -170,7 +170,9 @@ def get_sdfg_path(
             )
         # Check resolution per tile
         build_resolution = ast.literal_eval(build_info_file.readline())
-        if config.tile_resolution != build_resolution:
+        if (config.tile_resolution[0] / config.layout[0]) != (
+            build_resolution[0] / build_layout[0]
+        ):
             raise RuntimeError(
```
jdahm, line 176: This change is probably fine, but just curious - what's the rationale behind it? Was this an existing bug?

dsl/pace/dsl/dace/dace_config.py
```diff
@@ -7,10 +7,10 @@
...
 
 
-# TODO (floriand): Temporary deactivate the distributed compiled
-# until we deal with the Grid data inlining during orchestration
-# See github issue #301
-DEACTIVATE_DISTRIBUTED_DACE_COMPILE = True
+# This can be turned on to revert compilation for orchestration
+# in a rank-compile-itself more, instead of the distributed top-tile
+# mechanism.
+DEACTIVATE_DISTRIBUTED_DACE_COMPILE = False
```
jdahm, line 13: If this value changes in practice, please make sure this is called out in the first message and in the body of the squashed merge commit. Maybe it's already there, I can't tell.

dsl/pace/dsl/dace/orchestration.py
```diff
@@ -425,7 +425,7 @@ def orchestrate(
     obj: object,
     config: DaceConfig,
     method_to_orchestrate: str = "__call__",
-    dace_constant_args: List[str] = [],
+    dace_compiletime_args: List[str] = [],
```
jdahm, line 428: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

fv3core/pace/fv3core/stencils/dyn_core.py
```diff
@@ -998,7 +1011,8 @@ def __call__(
         if self._do_del2cubed:
             self._halo_updaters.heat_source.update()
             # TODO: move dependence on da_min into init of hyperdiffusion class
-            cd = constants.CNST_0P20 * self._da_min
+            da_min: float = self._get_da_min()
```
jdahm, line 1014: AFAIK the float type hint doesn't do anything here. 

```suggestion
            da_min = self._get_da_min()
```

fv3core/pace/fv3core/stencils/updatedzd.py
```diff
@@ -207,6 +208,10 @@ def __init__(
                 can be used as an approximation
             column_namelist: dictionary of parameter columns
         """
+        orchestrate(
+            obj=self,
+            config=stencil_factory.config.dace_config,
+        )
```
jdahm, line 214: It seems like every time `orchestrate` is called it uses keyword-only arguments, yet these are implemented as positional arguments in the definition of the function. Might I suggest switching the definition to,

```python
def orchestrate(*, obj, ...):
    ...
```

in `pace.dsl.dace.orchestration`.

fv3core/pace/fv3core/stencils/fv_dynamics.py
```diff
@@ -480,10 +484,11 @@ def _compute(self, state: DycoreState, timer: pace.util.Timer):
                         self._timestep,
                     )
                 if last_step:
+                    da_min: float = self._get_da_min()
```
jdahm, line 487: Does the type hint do anything?

```suggestion
                    da_min = self._get_da_min()
```

dsl/pace/dsl/dace/build.py
FlorianDeconinck: It was a bug due to a bad multi-branch merge. Basically tile_resolution is decomposition dependent (e.g. 6 ranks = 192, 54 ranks = 576) but what is inlined in dace is the loop sizes, e.g. 192. So you have to divide to see if those match

dsl/pace/dsl/dace/dace_config.py
FlorianDeconinck: Well this is really going back to the default - but keeping the flag for quick debug. I have it spelled out in the commit, not sure how far down

fv3core/pace/fv3core/stencils/fv_dynamics.py
```diff
@@ -531,7 +533,7 @@ def post_remap(
         self,
         state: DycoreState,
         is_root_rank: bool,
-        da_min: FloatFieldIJ,
+        da_min: float,
```
jdahm, line 536: Should this be float, or is it `pace.dsl.typing.Float`?

fv3core/pace/fv3core/stencils/dyn_core.py
FlorianDeconinck: It actually is needed so DaCe can parse the callback return value properly

fv3core/pace/fv3core/stencils/fv_dynamics.py
FlorianDeconinck: Yes. see above

fv3core/pace/fv3core/stencils/fv_dynamics.py
FlorianDeconinck: It's an actual float, per DampingCoefficient.da_min definition

dsl/pace/dsl/dace/orchestration.py
FlorianDeconinck: Good catch! I didn't remember that (I think I try to erase this from my mind it's so gross). Will change

fv3core/pace/fv3core/stencils/updatedzd.py
FlorianDeconinck: It really is just a copy-paste. I'd rather spell out the arguments in the definition. I dislike not having arguments laid out

jdahm: The arguments would still be laid out in the definition, the `*` in Python just forces the callers of the function to use keyword arguments for all remaining arguments.

jdahm: So, all your existing calls would remain correct (and would be the _only_ correct way to call it)

FlorianDeconinck: That might be a good idea then, since this is quite the weird function overall

dsl/pace/dsl/dace/build.py
```diff
@@ -170,7 +170,9 @@ def get_sdfg_path(
             )
         # Check resolution per tile
         build_resolution = ast.literal_eval(build_info_file.readline())
-        if config.tile_resolution != build_resolution:
+        if (config.tile_resolution[0] / config.layout[0]) != (
+            build_resolution[0] / build_layout[0]
```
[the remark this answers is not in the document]
elynnwu: Looks like Johann already commented on this too.

fv3core/pace/fv3core/stencils/divergence_damping.py
```diff
@@ -609,13 +639,20 @@ def __call__(
                 v_contra_dxc,
                 abs(dt),
             )
+
+        da_min: float = self._get_da_min()
+        if self._stretched_grid:
+            dd8 = da_min * self._d4_bg ** (self._nonzero_nord + 1)
+        else:
+            dd8 = (da_min_c * self._d4_bg) ** (self._nonzero_nord + 1)
```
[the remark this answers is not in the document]
FlorianDeconinck: This was in `init` but because `da_min` & `da_min_c` need to explicitly be given to `dace` as a runtime value, it cannot be in `init` anymore

dsl/pace/dsl/dace/build.py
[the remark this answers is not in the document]
FlorianDeconinck: This won't support != x/y, no. This fixes a bug that was reintroduced by a bad merge

elynnwu (approved):

PR #327, opened 2022-09-15
Fix caching of gt_cache directories on CircleCI

mcgibbon:
## Purpose

This PR fixes an issue where gt4py cache directories were not cached between executions on CircleCI.

## Infrastructure changes:

- Savepoint make targets now run `python3 -m gt4py.gt_src_manager install` before running tests inside docker, to avoid issues arising from git cloning these sources in a MPI-parallel context

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,7 +153,7 @@ jobs:
             TEST_ARGS="--backend=<<parameters.backend>> -v -s" DEV=y make savepoint_tests_mpi
           no_output_timeout: 3h
       - save_cache:
-          key: v12-gt_cache-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
+          key: v2-gt_cache-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
           paths:
             - .gt_cache
             - .gt_cache_000000
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ jobs:
           command: git submodule status external/gt4py | awk '{print $1;}' > gt4py_version.txt
       - restore_cache:
           keys:
-            - v2-gt_cache-serial-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
+            - v3-gt_cache-serial-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
             - v1-savepoints-{{ checksum "Makefile.data_download" }}
       - run:
           name: build image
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -10,9 +10,11 @@ RUN apt-get update && apt-get install -y make \
     netcdf-bin \
     libnetcdf-dev \
     python3 \
-    python3-pip
+    python3-pip \
```
mcgibbon, line 13: I will revert the changes to this file, they didn't help.

mcgibbon:
cpu mpi savepoint tests now pass in 5 minutes when fully cached, instead of 1.5h.

.circleci/config.yml
```diff
@@ -97,12 +99,13 @@ jobs:
...
           command: |
-            TEST_ARGS="--backend=<<parameters.backend>> -v -s" DEV=n make savepoint_tests
+            TEST_ARGS="--backend=<<parameters.backend>> -v -s" DEV=y make savepoint_tests
           no_output_timeout: 3h
       - save_cache:
-          key: v1-gt_cache-serial-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
+          key: v3-gt_cache-serial-<<parameters.backend>>-{{ checksum "gt4py_version.txt" }}
           paths:
             - .gt_cache
+            - .gt_cache_000000
```
elynnwu, line 108: I thought `GT_CACHE_ROOT` would pipe the cache directory?

Makefile
```diff
@@ -24,6 +24,7 @@ PULL ?=True
 DEV ?=y
 CHECK_CHANGED_SCRIPT=$(CWD)/changed_from_main.py
 CONTAINER_CMD?=docker
+SAVEPOINT_SETUP=pip3 list && python3 -m gt4py.gt_src_manager install
```
elynnwu, line 27: I think at some point this was part of the dockerfile? Also, I see that this is called during set up environment in circleci, does that not work?

mcgibbon:
jenkins fail run at https://jenkins.ginko.ch/job/pace_PR/2852/ relaunching to check if it's persistent or random

mcgibbon:
launch jenkins

Makefile
mcgibbon: The complicated case is that on the CircleCI test, we:
- Cannot run this outside of the container, because the environment is not set up / installed (I could add it just to run this command, but that feels like overkill)
- Are running the MPI-parallel tests, so if the process itself tries to clone these sources it happens 6 times in parallel, which leads to errors
- Need to bind-mount our directory into the container so that we can retain the caches, but doing so over-writes the gt4py directory (including gridtools sources) with our directory that does not have these

I could turn this into a `?=` and modify it only for the CircleCI tests, if that would be better? Normally this isn't an issue because we have/can get the gridtools sources on our local filesystem copy of gt4py.

.circleci/config.yml
mcgibbon: We don't set GT_CACHE_ROOT for this plan, though?

elynnwu: Right, looks like only for running 54 ranks mpi test

Makefile
elynnwu: Okay, just to double check, if we already had gridtools installed, this wouldn't do anything right?

PR #330, opened 2022-09-16
Fix origin and extent on lon/lat variables in MetricTerms

mcgibbon:
## Purpose

This PR fixes the origin and extent on lat, lon, lat_agrid, and lon_agrid attributes of MetricTerms.


elynnwu (approved):
lgtm

PR #333, opened 2022-09-19
Use raise_if_not_cached when RunMode.Run

### LINKED ISSUE #332: Use only lazy_stencil in Pace's FrozenStencil implementation
This currently uses both `stencil` as well as `lazy_stencil`, but the former can be eliminated.

jdahm:
## Purpose

Ensures stencils are compiled when in `RunMode.Run` and resolve #332.


dsl/pace/dsl/stencil.py
```diff
@@ -317,8 +317,6 @@ def __init__(
         )
         self.stencil_object: Optional[gt4py.StencilObject] = None
 
-        self._argument_names = tuple(inspect.getfullargspec(func).args)
-
```
jdahm, line 321: This exact line was run twice.

PR #339, opened 2022-09-20
Enable starting driver from Fortran restart files for TC case

ajdas1:
## Purpose

Allows a start from fortran restart files as needed to run the tropical cyclone initialization case.

## Code changes:

- Added GeneratedGridConfig and SerialboxGridConfig to pace.driver
- Added grid_config attribute to DriverConfig, used to select grid initialization approach instead of relying on the state initialization to do so
- Added FortranRestartConfig for state initialization
- Can now use fortran restart files to initialize the model (tc case)
- Possibility for grid transformation that allows for local refinement.
- removed unused "ks" variable from fv3core and dataclass containers
- added grid variables ee1, ee2, es1, and ew2 to GridData as they are needed by fv3core for baroclinic initialization
- deleted difficult-to-maintain mock-based driver tests, they are difficult to maintain and don't add much coverage past the driver tests which actually run the code
- removed unused stretch_grid attribute from Namelist class since Namelist is never used when defining grid initialization configuration
- fv3core grid dataclasses now use Quantity instead of gt4py storages

## Checklist

- [x] The names of all the new contributors have been added to CONTRIBUTORS.md


driver/pace/driver/driver.py
```diff
@@ -356,6 +368,125 @@ def exit_instead_of_build(self):
...
+        angle_data = pace.util.grid.AngleGridData.new_from_metric_terms(metric_terms)
+        grid_data = pace.util.grid.GridData(
+            horizontal_data=horizontal_data,
+            vertical_data=vertical_data,
+            contravariant_data=contravariant_data,
+            angle_data=angle_data,
+        )
+        self.state.grid_data = grid_data
+
+    def _transform_grid(self):
```
mcgibbon, line 459: Ideally these should exist on the grid class.

driver/pace/driver/gridconfig.py
```diff
@@ -1,23 +1,23 @@
...
-import numpy as np
-
 
 @dataclasses.dataclass
 class GridConfig:
-    stretch_mode: bool = False
-    stretch_factor: Optional[np.float_] = None
-    lon_target: Optional[np.float_] = None
-    lat_target: Optional[np.float_] = None
+    stretch_grid: bool = False
```
mcgibbon, line 7: If we're adding a class to configure the grid initialization, it should contain the code to initialize the grid. We should move the code out of the state-initialization classes into this. There would be two implementations - serialized grid, and generated grid.

driver/pace/driver/initialization.py
```diff
@@ -136,15 +136,25 @@ class RestartConfig(Initializer):
...
+            communicator.rank,
+            quantity_factory,
+            communicator,
+            self.fortran_data,
         )
+
+        # TODO
+        # follow what fortran does with restart data after reading it
+        # should eliminate small differences between restart input and
+        # serialized test data
```
mcgibbon, line 157: Is this TODO still relevant? If the directory is a pace restart directory, do we even want to do that?

driver/pace/driver/initialization.py
```diff
@@ -136,15 +136,25 @@ class RestartConfig(Initializer):
 
     path: str = "."
     start_time: datetime = datetime(2000, 1, 1)
+    fortran_data: bool = False
```
mcgibbon, line 139: Instead of having this flag, it should be possible to automatically look at the path and determine whether it's a fortran restart or pace restart directory. They look very different.

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+    state_dict = pace.util.open_restart(path, communicator, fortran_restart=True)
+
+    state = _dict_state_to_driver_state(state_dict, state, is_gpu_backend)
+
+    return state
+
+
+def _driver_state_to_dict(
+    driver_state: Union[fv3core.DycoreState, pace.physics.PhysicsState, TendencyState],
+    restart_file_prefix: str,
```
mcgibbon, line 211: Reading this function it is very unclear to me why the restart_file_prefix argument is used the way it is. I think instead you want to check `if isinstance(<some value>, pace.util.Quantity)` and remove this argument.

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+    """
+    Takes a Pace driver state
+    and returns a dict of state quantities with their Fortran names
+    """
+
+    dict_state = {}
+
+    for _field in fields(type(driver_state)):
+        if "units" in _field.metadata.keys():
+            if is_gpu_backend:
```
mcgibbon, line 223: why would this behavior depend on whether `is_gpu_backend`?

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+):
+    """
+    Takes a Pace driver state
+    and returns a dict of state quantities with their Fortran names
+    """
+
+    dict_state = {}
+
+    for _field in fields(type(driver_state)):
+        if "units" in _field.metadata.keys():
```
mcgibbon, line 222: Why does it matter whether the field has units metadata?

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+    is_gpu_backend: bool,
+):
+    """
+    Takes a Pace driver state
+    and returns a dict of state quantities with their Fortran names
+    """
+
+    dict_state = {}
+
+    for _field in fields(type(driver_state)):
```
mcgibbon, line 221: Instead of doing this dynamically, it would be much safer to have a manual list of field names to iterate over.

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+    is_gpu_backend: bool,
+):
+    """
+    Takes a dict of state quantities with their Fortran names and a driver state
+    and populates the driver state with quantities from the dict.
+    """
+
+    for _field in fields(type(driver_state)):
+        if "units" in _field.metadata.keys():
+            if is_gpu_backend:
```
mcgibbon, line 246: I think you could set `to_type` in an if-elif on is_gpu_backend and then have the code below only once, instead of copy-pasted twice.

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+    driver_state: Union[fv3core.DycoreState, pace.physics.PhysicsState, TendencyState],
+    is_gpu_backend: bool,
+):
+    """
+    Takes a dict of state quantities with their Fortran names and a driver state
+    and populates the driver state with quantities from the dict.
+    """
+
+    for _field in fields(type(driver_state)):
+        if "units" in _field.metadata.keys():
```
mcgibbon, line 245: It feels dangerous to have the restart code break if we remove units metadata from a dataclass field.

driver/pace/driver/state.py
```diff
@@ -189,24 +297,29 @@ def _restart_driver_state(
...
     physics_state = pace.physics.PhysicsState.init_zeros(
         quantity_factory=quantity_factory, active_packages=active_packages
     )
-    physics_state = _overwrite_state_from_restart(
-        path,
-        rank,
-        physics_state,
-        "restart_physics_state",
-        backend_uses_gpu,
-    )
```
mcgibbon, line 209: Why is this removed?

driver/pace/driver/state.py
```diff
@@ -175,12 +179,116 @@ def _overwrite_state_from_restart(
...
+
+    return driver_state
+
+
 def _restart_driver_state(
     path: str,
     rank: int,
     quantity_factory: pace.util.QuantityFactory,
     communicator: pace.util.CubedSphereCommunicator,
+    fortran_data: bool = False,
```
mcgibbon, line 289: I think you can delete this argument by making `_overwrite_state_from_restart` dynamically detect whether the path is a fortran or pace restart folder.

fv3core/pace/fv3core/_config.py
```diff
@@ -266,6 +266,7 @@ class DynamicalCoreConfig:
     fv_sg_adj: int = NamelistDefaults.fv_sg_adj
     n_sponge: int = NamelistDefaults.n_sponge
     namelist_override: Optional[str] = None
+    stretch_grid: bool = NamelistDefaults.stretch_grid
```
mcgibbon, line 269:
```suggestion
```
You moved this to GridConfig (which is good).

physics/tests/savepoint/translate/translate_driver.py
```diff
@@ -51,14 +53,13 @@ def compute_parallel(self, inputs, communicator):
...
-                "type": "predefined",
+                "type": "restart",
                 "config": {
-                    "dycore_state": dycore_state,
-                    "grid_data": self.grid.grid_data,
-                    "damping_coefficients": self.grid.damping_coefficients,
-                    "driver_grid_data": self.grid.driver_grid_data,
-                    "physics_state": physics_state,
-                    "tendency_state": tendency_state,
+                    "path": "/home/ajdas/pace/restart_data/v1.0",
```
mcgibbon, line 58: This will break the test.

tests/main/driver/test_restart_fortran.py
```diff
@@ -0,0 +1,248 @@
+import os
```
mcgibbon, line 1: I'm going to skip reviewing this file for now.

util/pace/util/_legacy_restart.py
```diff
@@ -24,6 +23,7 @@ def open_restart(
     only_names: Iterable[str] = None,
     to_state: dict = None,
     tracer_properties: RestartProperties = None,
+    fortran_restart: bool = False,
```
mcgibbon, line 26: What is the purpose of this argument? This function is specifically only for loading Fortran restarts (hence the filename).

util/pace/util/_legacy_restart.py
```diff
@@ -111,13 +119,13 @@ def _apply_dims(da, new_dims):
...
     for name, da in state.items():
-        if name in restart_properties.keys():
-            properties = restart_properties[name]
-            new_dims = properties["dims"]
-            new_state[name] = _apply_dims(da, new_dims)
-            new_state[name].attrs["units"] = properties["units"]
-        else:
-            new_state[name] = copy.deepcopy(da)
+        for key in restart_properties.keys():
+            if name in restart_properties[key]["driver_name"]:
```
mcgibbon, line 123: Why was this for-if flipped? Or, why was this function changed?

util/pace/util/_legacy_restart.py
```diff
@@ -141,20 +149,27 @@ def prepend_label(filename, label=None):
...
         name: value
         for name, value in state.items()
         if ((name == "time") or ("units" in value.attrs)) and name in only_names
     }
+
     for name, array in state.items():
         if name != "time":
             array.load()
             state[name] = Quantity.from_data_array(array)
+
```
mcgibbon, line 172: nit: Please remove these newlines.

util/pace/util/_legacy_restart.py
```diff
@@ -164,6 +179,7 @@ def _get_restart_standard_names(restart_properties: RestartProperties = None):
     if restart_properties is None:
         restart_properties = RESTART_PROPERTIES
     return_dict = {}
-    for std_name, properties in restart_properties.items():
-        return_dict[properties["restart_name"]] = std_name
+    for _, properties in restart_properties.items():
+        return_dict[properties["restart_name"]] = properties["driver_name"]
```
mcgibbon, line 183: This function is used by the Python wrapper and the ML team, we can't edit it in this way. Please revert this change. We can chat about what you need to do here and where to do it.

From what I can tell, you could use the existing functions as-is to load a Fortran restart state into a dictionary, and then in that function rename the variables that have different names on the driver state. The "safe" way to do that is to manually have many lines like
```
driver_state.dycore_state.u = restart["u"]
```
since code introspection can detect these accesses and e.g. switch it to 
```
driver_state.dycore_state.x_wind = restart["u"]
```
automatically if we ever renamed that variable. This is a lot easier to do with Copilot though.

util/pace/util/_properties.py
```diff
@@ -16,437 +16,614 @@
     "accumulated_x_courant_number": {
         "dims": [Z_DIM, Y_DIM, X_DIM],
         "restart_name": "cx",
+        "driver_name": "cx",
```
mcgibbon, line 19: Please revert the changes to this file.

util/pace/util/_properties.py
```diff
@@ -16,437 +16,614 @@
...
+        "units": "kg/kg",
+        "restart_file": "fv_tracer.res.tile",
+    },
+    "vertical_thickness_of_atmospheric_layer": {
+        "dims": [Z_DIM, Y_DIM, X_DIM],
+        "restart_name": "DZ",
+        "driver_name": "delz",
+        "units": "m",
+        "restart_file": "fv_core.res.tile",
+    },
```
mcgibbon, line 628: In the fortran restart loading code, these are treated as "tracer properties", since they're defined at runtime by the tracer configuration file. These aren't statically-defined.

util/pace/util/grid/stretch_transformation.py
```diff
@@ -22,66 +27,73 @@ def direct_transform(
     centeres on lon_target, lat_target.
 
     Args:
-        lon (in) in radians
-        lat (in) in radians
-        stretch_factor (in) stretch_factor (e.g. 3.0 means that the resolution on
-            tile 6 becomes 3 times as fine)
+        lon (in) in radians # Quantity or ndarray (cupy or numpy)
+        lat (in) in radians # Quantity or ndarray (cupy or numpy)
```
mcgibbon, line 31: This is already stated in the type hint. Can you remove it below too?

```suggestion
        lon (in) in radians
        lat (in) in radians
```

util/pace/util/grid/stretch_transformation.py
```diff
@@ -22,66 +27,73 @@ def direct_transform(
...
+        stretch_factor (in) stretch_factor (e.g. 3.0 means that the resolution
+            on tile 6 becomes 3 times as fine)
         lon_target (in) in degrees (from namelist)
         lat_target (in) in degrees (from namelist)
```
mcgibbon, line 35: please add numpy_module to the docstring

util/pace/util/grid/stretch_transformation.py
```diff
@@ -1,19 +1,24 @@
...
     *,
-    lon: Union[Quantity, np.ndarray],
-    lat: Union[Quantity, np.ndarray],
+    lon,
+    lat,
     stretch_factor: np.float_,
     lon_target: np.float_,
     lat_target: np.float_,
-) -> Tuple[Union[np.ndarray, Quantity], Union[np.ndarray, Quantity]]:
+    numpy_module,
```
mcgibbon, line 20: nit: Please call this argument `np` as we do in the grid generation (e.g. util/pace/util/grid/gnomonic.py) in this and the other functions.

util/pace/util/grid/stretch_transformation.py
```diff
@@ -90,14 +102,14 @@ def direct_transform(
...
-    elif isinstance(lon, np.ndarray):
+    elif isinstance(lon, numpy_module.ndarray):
         lon_transform = lon_trans
         lat_transform = lat_trans
 
     return lon_transform, lat_transform
 
 
-def _sign(input, pn):
+def _sign(input, pn, numpy_module):
```
mcgibbon, line 112: I'm kind of confused by the existence of this function. Isn't there a `np.sign` and `cp.sign`?

mcgibbon: Wait does this actually do `np.abs(input) * np.sign(pn)`? Can you use that instead? Why is it called `pn`?

mcgibbon (changes requested):

util/pace/util/_properties.py
mcgibbon: Because we hard-code the tracer configuration in everything we do currently in Pace, you can define a static/constant TRACER_PROPERTIES dictionary somewhere inside `pace.driver` (possibly in the restart initialization code) which we could later generate from the microphysics configuration.

util/pace/util/grid/stretch_transformation.py
ajdas1: Wasn't aware of np.sign, that's now done.

driver/pace/driver/state.py
ajdas1: We fill in dycore state from restart data. Elynn said that the physics state gets filled in appropriately from dycore, so no need to write into it

driver/pace/driver/state.py
elynnwu: I think this might be a coping issue? This was not touched in the [PR](https://github.com/ai2cm/pace/pull/329/) to get things to run on gpu. I think the only thing you need here is just:
`dict_state[_field.name] = driver_state.__dict__[_field.name].data[:]`

driver/pace/driver/state.py
elynnwu: I believe I started the unit thing because not all dycore state and physics state need to be filled. Dycore state has two float variables, physics state doesn't use Quantity but do have units in the metadata. So it was convenient to check whether units exist. I remember also having issues with `isinstance` checking physics state `FloatField`. Alternatively, we could also just check if it's not `bdt` or `mdt`.

driver/pace/driver/state.py
elynnwu: This is also a good idea.

driver/pace/driver/state.py
elynnwu: Yes, I realized that `copy_dycore_to_physics` means physics state always gets reset from dycore state. So we do not even need to write out physics restart in the pace restart format.

driver/examples/configs/baroclinic_c12_read_restart_fortran.yml
```diff
@@ -0,0 +1,93 @@
...
+  compilation_config:
+    backend: numpy
+    rebuild: false
+    validate_args: true
+    format_source: false
+    device_sync: false
+initialization:
+  config:
+    path: /home/ajdas/pace/restart_data/v1.0
+    start_time: 2016-08-11 00:00:00
```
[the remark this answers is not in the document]
ajdas1: Added function to fortran restart config

driver/pace/driver/initialization.py
ajdas1: Yes, I think so -- but it now only exists in the fortran restart config

util/pace/util/_legacy_restart.py
ajdas1: Removed

util/pace/util/_legacy_restart.py
ajdas1: restart_properties.keys() are things like `x_wind`, `x_wind_on_c_grid`, and would not match the names in `state.items()`

driver/examples/configs/baroclinic_c12_read_restart_fortran.yml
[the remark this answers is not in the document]
mcgibbon: You still need to update this config (also the data path).

driver/examples/configs/baroclinic_c12_read_restart_fortran.yml
```diff
@@ -0,0 +1,93 @@
...
+    fortran_data: true
```
mcgibbon, line 12: Also need to remove fortran_data setting and update the initialization type.

driver/examples/configs/tropical_read_restart_fortran.yml
```diff
@@ -0,0 +1,99 @@
...
+    rebuild: false
+    validate_args: true
+    format_source: false
+    device_sync: false
+initialization:
+  config:
+    path: /home/ajdas/pace/restart_data/v5.0
+    start_time: 2020-08-26 12:00:01
+    fortran_data: true
+  type: restart
```
mcgibbon, line 15: Needs updating.

driver/pace/driver/driver.py
```diff
@@ -287,10 +295,14 @@ def exit_instead_of_build(self):
                 communicator=communicator,
                 stencil_compare_comm=stencil_compare_comm,
             )
-
             self.state = self.config.initialization.get_driver_state(
                 quantity_factory=self.quantity_factory, communicator=communicator
             )
+
+            if self.config.initialization.config.fortran_data is True:
```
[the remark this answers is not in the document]
mcgibbon: There's still code here, needs to be resolved.

mcgibbon (commented):
Might need to push your changes.

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,273 @@
...
+        lat_target: desired center latitude for refined tile (deg)
+        tc_ks: something to do with friction in the vertical ???Ajda
+        restart_path: path to restart data
+    """
+
+    stretch_grid: bool = False
+    stretch_factor: Optional[float] = None
+    lon_target: Optional[float] = None
+    lat_target: Optional[float] = None
+    ks: int = 0
```
mcgibbon, line 96: probably ignore me: It would be cool if stretch_grid wasn't an option, but instead stretch_factor is by default 1.0 (or 0.0, idk which but no stretching) and lon/lat_target are by default the south pole, so that by default no stretching is done but there isn't an option to "turn on" the stretching.

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,273 @@
...
+
+@GridInitializerSelector.register("serialbox")
+@dataclasses.dataclass
+class SerialboxConfig(GridInitializer):
+    """
+    Configuration for Serialbox initialization.
+    """
+
+    path: str
+    serialized_grid: bool
```
mcgibbon, line 145: You can remove the `serialized_grid` option, if we're using the SerialboxConfig to initialize the grid then we always want the serialized grid.

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,273 @@
...
+    grid.data[:, :, 0] = lon_transform[:]
+    grid.data[:, :, 1] = lat_transform[:]
+
+    metric_terms._grid.data[:] = grid.data[:]
+    metric_terms._init_agrid()
+
+    return metric_terms
+
+
+def _replace_vertical_grid(self, metric_terms) -> GridData:
```
mcgibbon, line 233: This only uses `ks` and `restart_path` from `self`, you should probably pass those directly instead.

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,273 @@
...
+    Replaces the vertical grid generators from metric terms (ak and bk) with
+    their fortran restart values (in fv_core.res.nc).
+    Then re-generates grid data with the new vertical inputs.
+    p_ref(?)
+    """
+
+    restart_files = os.listdir(self.restart_path)
+    data_file = [fl for fl in restart_files if "fv_core.res.nc" in fl][0]
+
+    ak_bk_data_file = self.restart_path + data_file
```
mcgibbon, line 244: Use `pathlib.Path(self.restart_path) / data_file` instead, it will work on all platforms and deal with missing/extra "/".

driver/pace/driver/gridconfig.py
```diff
@@ -1,23 +1,23 @@
 import dataclasses
 from typing import Optional
 
-import numpy as np
-
 
 @dataclasses.dataclass
 class GridConfig:
```
mcgibbon, line 6: You should be able to delete this file, its functionality was moved to `grid.py`.

driver/pace/driver/initialization.py
```diff
@@ -95,15 +105,15 @@ def get_driver_state(
         self,
         quantity_factory: pace.util.QuantityFactory,
         communicator: pace.util.CubedSphereCommunicator,
+        damping_coefficients: pace.util.grid.DampingCoefficients,
+        driver_grid_data: pace.util.grid.DriverGridData,
+        grid_data: pace.util.grid.GridData,
     ) -> DriverState:
+        # Ajda
+        # do I need to keep metric terms here?
```
mcgibbon, line 113: This is a bug, not your fault though.

You need to refactor `baroclinic_init.init_baroclinic_state` so it takes `GridData` instead of taking `MetricTerms`, and then remove the `metric_terms` here. Currently I think if you use a serialized grid, the state initialization here will completely ignore it and use a generated grid instead for initializing the model state.

driver/pace/driver/initialization.py
```diff
@@ -141,13 +151,96 @@ def get_driver_state(
...
+
+        # TODO
+        # follow what fortran does with restart data after reading it
+        # should eliminate small differences between restart input and
+        # serialized test data
+
+        return state
+
+
+def _update_fortran_restart_pe_peln(state):
```
mcgibbon, line 224: Can you add a short docstring describing what this is dong and why? And type hint the input?

driver/pace/driver/initialization.py
```diff
@@ -141,13 +151,96 @@ def get_driver_state(
...
+    pe = state.dycore_state.pe
+    peln = state.dycore_state.peln
+    delp = state.dycore_state.delp
+
+    np = state.dycore_state.u.np
+
+    for level in range(pe.data.shape[2]):
+        pe.data[:, :, level] = ptop + np.sum(delp.data[:, :, :level], 2)
+
+    peln.data[:] = np.log(pe.data[:])
```
mcgibbon, line 236: it's unlikely but it's technically possible for different quantities to have different np (i.e. live on cpu or gpu independently).
```suggestion
    for level in range(pe.data.shape[2]):
        pe.data[:, :, level] = ptop + delp.np.sum(delp.data[:, :, :level], 2)

    peln.data[:] = pe.np.log(pe.data[:])
```

driver/pace/driver/initialization.py
```diff
@@ -141,13 +151,96 @@ def get_driver_state(
...
+
+    state.dycore_state.pe = pe
+    state.dycore_state.peln = peln
+
+    return state
```
mcgibbon, line 241: These lines do nothing, as pe and peln are already the ones attached to dycore_state.

```suggestion
    return state
```

driver/pace/driver/state.py
```diff
@@ -95,8 +96,26 @@ def load_state_from_restart(
...
+            quantity_factory=quantity_factory,
+            communicator=communicator,
+        )
+        # state = _restart_driver_state(
+        #     restart_path,
+        #     communicator.rank,
+        #     quantity_factory,
+        #     communicator,
+        #     fortran_data,
+        # )
```
mcgibbon, line 110:
```suggestion
        )
```

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+    communicator: pace.util.CubedSphereCommunicator,
+    state: Union[fv3core.DycoreState, pace.physics.PhysicsState, TendencyState],
+    is_gpu_backend: bool,
+):
+    """
+    Args:
+        path: path to restart files
+        rank: current rank number
+        state: an empty state
+        restart_file_prefix: file prefix name to read
```
mcgibbon, line 208: Docstring needs updating.

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+    # Ajda
+    # at this point the data for every field in dict_state is 0.
+    return dict_state
+
+
+def _dict_state_to_driver_state(
+    fortran_state: dict,
+    driver_state: Union[fv3core.DycoreState, pace.physics.PhysicsState, TendencyState],
+    is_gpu_backend: bool,
+):
```
mcgibbon, line 246: Because this modifies driver_state in-place, please modify it so it doesn't return driver_state. This gives the impression you're returning a new driver_state instead of modifying in-place.

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+    """
+    Args:
+        path: path to restart files
+        rank: current rank number
+        state: an empty state
+        restart_file_prefix: file prefix name to read
+
+    Returns:
+        state: new state filled with restart files
```
mcgibbon, line 211: Does this return a new state? It looks like it modifies the state passed in. If it modifies the state passed in, please also modify this so it doesn't return `state`.

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+    """
+    state_dict = _driver_state_to_dict(state)
+
+    state_dict = pace.util.open_restart(path, communicator)
```
mcgibbon, line 215: state_dict is overwritten without ever being used.
```suggestion
    state_dict = pace.util.open_restart(path, communicator)
```

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+    dict_state = {}
+
+    for field in fortran_restart_to_pace_dict.keys():
+        dict_state[fortran_restart_to_pace_dict[field]] = driver_state.__dict__[
+            field
+        ].data[:]
+
+    # Ajda
+    # at this point the data for every field in dict_state is 0.
+    return dict_state
```
mcgibbon, line 239: I think you can delete this function.

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
+
+    for field in fortran_restart_to_pace_dict.keys():
+        driver_state.__dict__[field].view[:] = np.transpose(fortran_state[field].data)
+
+        if is_gpu_backend:
+            # driver_state.__dict__[field].view[:] = gt_utils.asarray(
+            #     np.transpose(fortran_state[field].data), to_type=cp.ndarray,
+            # )
+            # Ajda
+            # not sure if this will work?? Internet told me cupy has transpose
```
mcgibbon, line 260: Is there any way to find out before merging? @elynnwu we might want to add GPU-backend coverage of or non-savepoint (main) tests.

driver/pace/driver/state.py
```diff
@@ -175,38 +194,160 @@ def _overwrite_state_from_restart(
...
-        dycore_state,
-        "restart_dycore_state",
-        backend_uses_gpu,
-    )
+
+    is_fortran_restart = False
+    restart_files = os.listdir(path)
+    for fl in restart_files:
+        if "fv_core.res.nc" in fl:
+            is_fortran_restart = True
```
mcgibbon, line 328:
```suggestion
    restart_files = os.listdir(path)
    is_fortran_restart = any(fname.endswith("fv_core.res.nc") for fname in restart_files)
```

tests/main/driver/test_restart_fortran.py
mcgibbon: This test file has a lot going on, and I'm not sure what it's testing. I think you started based on the `test_restart_serial` but what it needs to test is very different from what you need to test (it's mostly testing the fact that we write restarts).

Instead I'd suggest writing a unit test that:
- initializes a FortranRestartConfig referencing the data used in the util tests (you can use `os.path.dirname(os.path.abspath(__file__))` to get the directory of this test file itself, and then use relative paths from there)
- uses that initialized config to generate a driver state (use a generated grid)
- compares one variable for each staggering type against data in the restart netCDF file to ensure data got loaded with the right alignment. Probably u/v/temperature are enough to compare.

Include any other assertions or cases you think are important, e.g. whether pe/peln are overwritten correctly.

util/pace/util/_legacy_restart.py
```diff
@@ -38,29 +42,43 @@ def open_restart(
...
     """
+
     if tracer_properties is None:
         restart_properties = RESTART_PROPERTIES
+        restart_properties.update(extra_restart_properties)  # add microphysics
+        # original restart properties don't have that stuff
     else:
         restart_properties = {**tracer_properties, **RESTART_PROPERTIES}
+        restart_properties.update(extra_restart_properties)
+
```
mcgibbon, line 53:
```suggestion
    else:
        restart_properties = {**tracer_properties, **RESTART_PROPERTIES}
    restart_properties.update(extra_restart_properties)  # add microphysics
    # original restart properties don't have that stuff

```

util/pace/util/_legacy_restart.py
```diff
@@ -38,29 +42,43 @@ def open_restart(
...
     rank = communicator.rank
     tile_index = communicator.partitioner.tile_index(rank)
     state = {}
     if communicator.tile.rank == constants.ROOT_RANK:
         for file in restart_files(dirname, tile_index, label):
+
```
mcgibbon, line 59:
```suggestion
```

util/pace/util/_legacy_restart.py
```diff
@@ -38,29 +42,43 @@ def open_restart(
...
+    # if not fortran_restart:
+    #     if to_state is None:
+    #         state = communicator.tile.scatter_state(state)
+    #     else:
+    #         state = communicator.tile.scatter_state(state, recv_state=to_state)
+    # Ajda
+    # fortran restart files don't need to be scattered as each tile has its own
+    # file ... At least when it's run on 6 ranks.
+    # Not sure what happens if you have more (do you get more restart files?) ...
```
mcgibbon, line 80: Please restore this code. We need to be able to load restarts when we have more than 6 ranks.

```suggestion
    # if not fortran_restart:
    #     if to_state is None:
    #         state = communicator.tile.scatter_state(state)
    #     else:
    #         state = communicator.tile.scatter_state(state, recv_state=to_state)
    # Ajda
    # fortran restart files don't need to be scattered as each tile has its own
    # file ... At least when it's run on 6 ranks.
    # Not sure what happens if you have more (do you get more restart files?) ...
```

util/pace/util/initialization/__init__.py
```diff
@@ -1,2 +1,6 @@
 from .allocator import QuantityFactory
+from .fortran_restart_to_pace_variable_names import (
+    extra_restart_properties,
+    fortran_restart_to_pace_dict,
```
mcgibbon, line 4: I don't see this file committed, but it belongs in `pace.driver`. Can you put this data there instead?

util/pace/util/_legacy_restart.py
```diff
@@ -1,7 +1,11 @@
-import copy
 import os
 from typing import BinaryIO, Generator, Iterable
 
+from pace.util.initialization import (
```
mcgibbon, line 4: I'm concerned about any changes to this file. Are you able to write this so that this file is exactly the same as on main, but instead you have code in `pace.driver.initialization` that e.g. assigns variables to the correct driver state variable from the right fortran name?

util/pace/util/grid/helper.py
```diff
@@ -112,11 +112,10 @@ class VerticalGridData:
     """
 
     # TODO: make these non-optional, make FloatFieldK a true type and use it
-    ptop: float
     ks: int
     ak: Optional[Any] = None
     bk: Optional[Any] = None
-    p_ref: Optional[Any] = None
+    # p_ref: Optional[Any] = None
```
mcgibbon, line 118: Delete commented code in this file.

util/pace/util/namelist.py
```diff
@@ -105,6 +105,7 @@ class NamelistDefaults:
     tice = 273.16  # set tice = 165. to turn off ice - phase phys (kessler emulator)
     alin = 842.0  # "a" in lin1983
     clin = 4.8  # "c" in lin 1983, 4.8 -- > 6. (to ehance ql -- > qs)
+    stretch_grid = False
```
mcgibbon, line 108: Does this actually need to be added to our Namelist class? It's a little misleading to put it here, because the initialization code doesn't read from the namelist at all. I think you removed the place where the dycore was accessing this by putting it on the grid data containers instead.

mcgibbon (changes requested):

driver/examples/configs/tropical_read_restart_fortran.yml
```diff
@@ -0,0 +1,99 @@
...
+    path: /home/ajdas/pace/restart_data/v5.0/
+  type: fortran_restart
+grid_init:
+  config:
+    stretch_grid: true
+    stretch_factor: 3.0
+    lon_target: 172.5
+    lat_target: 17.5
+    ks: 0
+    restart_path: /home/ajdas/pace/restart_data/v5.0/
```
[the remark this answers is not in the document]
ajdas1: Now downloads data to predefined directory, and reads restart from there if `vertical_grid_from_restart` is True

util/pace/util/_legacy_restart.py
ajdas1: I didn't really touch this at all - this was all Oliver :D

util/pace/util/_legacy_restart.py
ajdas1: I think this is resolved? maybe?

driver/pace/driver/initialization.py
ajdas1: I refactored, but for now left the original stuff as comments, in case something breaks. Needs to be tested.

.gitignore
```diff
@@ -162,3 +162,5 @@ RESTART/
 _dacegraphs
 
 **/testing/output/*
+*.png
+*.nc
```
elynnwu, line 166: Do we actually want to ignore all png and nc files?

driver/pace/driver/diagnostics.py
```diff
@@ -135,19 +132,11 @@ def _get_derived_state(self, state: DriverState):
...
-                dims=("x_interface", "y_interface"),
-                origin=metadata.origin,
-                extent=(metadata.extent[0] + 1, metadata.extent[1] + 1),
-                units="rad",
-            )
-            zarr_grid[name] = grid_quantity
+    def store_grid(self, grid_data: pace.util.grid.GridData):
+        zarr_grid = {
+            "lat": grid_data.lat,
+            "lon": grid_data.lon,
```
elynnwu, line 138: 👏

driver/pace/driver/driver.py
```diff
@@ -201,6 +290,84 @@ def from_dict(cls, kwargs: Dict[str, Any]) -> "DriverConfig":
...
+            config_dict["physics_config"].pop(field, None)
+        config_dict["initialization"]["type"] = "restart"
+        config_dict["initialization"]["config"]["start_time"] = time
+        config_dict["initialization"]["config"]["path"] = restart_path
+        with open(f"{restart_path}/restart.yaml", "w") as file:
+            yaml.safe_dump(config_dict, file)
+
+
+@dataclasses.dataclass()
+class RestartConfig:
```
elynnwu, line 327: What's the reason behind this being inside `driver.py`?

driver/pace/driver/driver.py
```diff
@@ -482,23 +642,15 @@ def _write_performance_json_output(self):
...
-            comm=self.comm,
-            restart_path=restart_path,
-        )
-        self.restart.write_restart_config(
             comm=self.comm,
             time=self.time,
             driver_config=self.config,
-            restart_path=restart_path,
+            restart_path="RESTART",
         )
```
elynnwu, line 653: It looks to me that some of these don't need to be part of dace call back, not sure if this changes the behavior of dace.

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,220 @@
...
+    def from_dict(cls, config: dict):
+        instance = cls.registry.from_dict(config)
+        return cls(config=instance, type=config["type"])
+
+
+@GridInitializerSelector.register("generated")
+@dataclasses.dataclass
+class GeneratedGridConfig(GridInitializer):
+    """
+    Configuration for baroclinic initialization.
```
elynnwu, line 84:
```suggestion
    Grid calculated using metric terms 
```

driver/pace/driver/grid.py
```diff
@@ -0,0 +1,220 @@
...
+        driver_grid_data = DriverGridData.new_from_metric_terms(metric_terms)
+
+        return damping_coefficients, driver_grid_data, grid_data
+
+
+@GridInitializerSelector.register("serialbox")
+@dataclasses.dataclass
+class SerialboxGridConfig(GridInitializer):
+    """
+    Configuration for Serialbox initialization.
```
elynnwu, line 136:
```suggestion
    Grid read from Serialbox data
```

driver/pace/driver/initialization.py
```diff
@@ -122,14 +132,14 @@ def get_driver_state(
...
             driver_grid_data=driver_grid_data,
         )
 
 
 @InitializerSelector.register("restart")
 @dataclasses.dataclass
-class RestartConfig(Initializer):
+class RestartInit(Initializer):
     """
     Configuration for restart initialization.
```
elynnwu, line 144:
```suggestion
    Configuration for pace restart initialization.
```

driver/pace/driver/initialization.py
```diff
@@ -141,16 +151,83 @@ def get_driver_state(
...
+
+    @property
+    def start_time(self) -> datetime:
+        """Reads the last line in coupler.res to find the restart time"""
+        restart_files = os.listdir(self.path)
+
+        coupler_file = restart_files[
+            [fname.endswith("coupler.res") for fname in restart_files].index(True)
+        ]
+        # coupler_file = [fl for fl in restart_files if "coupler.res" in fl][0]
```
elynnwu, line 188: Delete this

driver/pace/driver/initialization.py
```diff
@@ -321,3 +363,31 @@ def get_driver_state(
...
+
+    for level in range(pe.data.shape[2]):
+        pe.data[:, :, level] = ptop + delp.np.sum(delp.data[:, :, :level], 2)
+
+    peln.data[:] = pe.np.log(pe.data[:])
+
+    state.dycore_state.pe = pe
+    state.dycore_state.peln = peln
+
+    return None
```
elynnwu, line 393: Remove return statement

driver/pace/driver/state.py
```diff
@@ -144,69 +147,61 @@ def save_state(self, comm, restart_path: str = "RESTART"):
...
-        path,
-        rank,
-        dycore_state,
-        "restart_dycore_state",
-        backend_uses_gpu,
+    fs = pace.util.get_fs(path)
+
+    is_fortran_restart = False
+    restart_files = fs.ls(path)
+    is_fortran_restart = any(
```
elynnwu, line 183: Do we need to default to False first?

stencils/pace/stencils/testing/grid.py
```diff
@@ -482,57 +483,324 @@ def grid_data(self) -> "GridData":
...
+            # es1=self.es1,
+            # ew2=self.ew2,
+            # a11=self.a11,
+            # a12=self.a12,
+            # a21=self.a21,
+            # a22=self.a22,
+            # edge_w=self.edge_w,
+            # edge_e=self.edge_e,
+            # edge_s=self.edge_s,
+            # edge_n=self.edge_n,
```
elynnwu, line 517: Delete this

stencils/pace/stencils/testing/grid.py
```diff
@@ -482,57 +483,324 @@ def grid_data(self) -> "GridData":
...
+            # self.cosa,
+            # self.cosa_u,
+            # self.cosa_v,
+            # self.cosa_s,
+            # self.sina_u,
+            # self.sina_v,
+            # self.rsina,
+            # self.rsin_u,
+            # self.rsin_v,
+            # self.rsin2,
```
elynnwu, line 703: delete this

stencils/pace/stencils/testing/grid.py
```diff
@@ -482,57 +483,324 @@ def grid_data(self) -> "GridData":
...
-            self.cos_sg3,
-            self.cos_sg4,
+            # self.sin_sg1,
+            # self.sin_sg2,
+            # self.sin_sg3,
+            # self.sin_sg4,
+            # self.cos_sg1,
+            # self.cos_sg2,
+            # self.cos_sg3,
+            # self.cos_sg4,
```
elynnwu, line 763: delete this

tests/main/driver/test_restart_fortran.py
```diff
@@ -0,0 +1,64 @@
...
+        nz=63,
+        n_halo=3,
+        extra_dim_lengths={},
+        layout=layout,
+        tile_partitioner=partitioner.tile,
+        tile_rank=0,
+    )
+
+    quantity_factory = QuantityFactory.from_backend(sizer=sizer, backend="numpy")
+    restart_dir = os.path.join(DIR, "../../../util/tests/data/c12_restart")
```
elynnwu, line 43: Would it be better to define pace dir at the top and use that throughout this file?

mcgibbon: Updated.

.gitignore
elynnwu: I'm okay with leaving it like this, just wanted to be careful when we do actually want to commit a png/nc file.

driver/pace/driver/driver.py
elynnwu: Fair enough.

elynnwu (approved):

mcgibbon (approved):

PR #350, opened 2022-10-04
Optimize column integral calculation 

elynnwu:
## Purpose

This PR makes the calculation of column integrated tracers faster.

## Code changes:

- diagnostics: use quantity np sum instead of native python sum

## Checklist

### DIFF
--- a/driver/pace/driver/diagnostics.py
+++ b/driver/pace/driver/diagnostics.py
@@ -4,8 +4,6 @@ import warnings
 from datetime import datetime, timedelta
 from typing import List, Optional, Union
 
-from sympy import ShapeError
-
 import pace.driver
 import pace.dsl
 import pace.stencils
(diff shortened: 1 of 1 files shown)


driver/pace/driver/diagnostics.py
```diff
@@ -177,7 +177,7 @@ def _compute_column_integral(
     if len(q_in.shape) < 3:
         assert ShapeError(f"{name} does not have vertical levels.")
     column_integral = pace.util.Quantity(
-        sum(
+        q_in.np.sum(
             q_in.data[:, :, k] * delp.data[:, :, k]
             for k in range(q_in.metadata.extent[2])
         ),
```
mcgibbon, line 183: I think it will only be faster if you do something like
```suggestion
        q_in.np.sum(
            q_in.data[:, :, q_in.origin[2]:q_in.origin[2] +q_in.extent[2]] * delp.data[:, :, q_in.origin[2]:q_in.origin[2] +q_in.extent[2]]
        ),
```
You should consider putting the weird slice thing in a `k_slice = slice(q_in.origin[2], q_in.origin[2] + q_in.extent[2]` object and adding a `if q_in.dims[2] != Z_DIM: raise NotImplementedError("this function assumes the z-dimension is the third dimension")`

driver/pace/driver/diagnostics.py
```diff
@@ -174,13 +172,13 @@ def _compute_column_integral(
         q_in: tracer mixing ratio
         delp: pressure thickness of atmospheric layer
     """
-    if len(q_in.shape) < 3:
-        assert ShapeError(f"{name} does not have vertical levels.")
+    if q_in.dims[2] != pace.util.Z_DIM:
+        raise NotImplementedError(
+            "this function assumes the z-dimension is the third dimension"
+        )
```
jdahm, line 178: Does `q_in` always have 3 dims so that indexing `[2]` directly does not raise an `IndexError`? Maybe make this
```python
assert len(q_in.dims) > 2
if q_in.dims[2] != pace.util.Z_DIM:
```

but maybe that's already assumed throughout the code.

elynnwu: updated

jdahm (approved):
Looks great!

PR #354, opened 2022-10-11
Remove _get_backend_opts from StencilConfig

jdahm:
Also cleanup an inspect call that was accidentally run twice.

### DIFF
--- a/dsl/pace/dsl/stencil_config.py
+++ b/dsl/pace/dsl/stencil_config.py
@@ -1,7 +1,6 @@
 import dataclasses
 import enum
 import hashlib
-import re
 from typing import Any, Callable, Dict, Hashable, Iterable, Optional, Sequence, Tuple
 
 from gtc.passes.oir_pipeline import DefaultPipeline, OirPipeline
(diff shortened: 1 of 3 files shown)


dsl/pace/dsl/stencil_config.py
```diff
@@ -216,36 +217,6 @@ def __eq__(self, other):
...
-            if "backend" not in option or using_option_backend:
-                backend_opts[name] = option["value"]
-
-        if device_sync is not None:
-            backend_opts["device_sync"] = device_sync
-        if format_source is not None:
-            backend_opts["format_source"] = format_source
-
-        return backend_opts
-
```
jdahm, line 248: This was a really confusing function, and we didn't end up ever using the confusing logic here, since we always passed `device_sync` and `format_source` anyway. I replaced this with a simple constant dict that this would have returned.

dsl/pace/dsl/stencil_config.py
```diff
@@ -172,11 +171,13 @@ class StencilConfig(Hashable):
     dace_config: Optional[DaceConfig] = None
 
     def __post_init__(self):
-        self.backend_opts = self._get_backend_opts(
-            self.compilation_config.device_sync, self.compilation_config.format_source
-        )
+        self.backend_opts = {
+            "device_sync": self.compilation_config.device_sync,
+            "format_source": self.compilation_config.format_source,
+        }
```
elynnwu, line 177: This is a good idea. Though I'm a little confused now from looking at `_get_backend_opts`, is there an expected value for the gpu backend? It looks like we've been running with `device_sync=True` for our runs lately.

elynnwu (approved):

dsl/pace/dsl/stencil_config.py
jdahm: Question for myself: does this affect the `dace:*` backends. The runs so far seem to have used `device_sync=True`. We think this should not be a namelist option.

jdahm: > does this affect the dace:* backends?

It adds a `device_sync` call the way other GPU backends do.

PR #362, opened 2022-10-19
Performance test uses driver

elynnwu:
## Purpose

This PR adds a script to do performance test from driver. 

## Code changes:

- Added `driver_performance.sh` to collect performance json file

## Infrastructure changes:

- pace performance test should now use driver instead of dynamics.py

## Checklist

- [ ] You have followed the coding standards guidelines established at [Code Review Checklist](https://drive.google.com/file/d/1R0nqOxfYnzaSdoYdt8yjx5J482ETI2Ft/view?usp=sharing).


driver/pace/driver/driver.py
```diff
@@ -598,11 +599,16 @@ def _critical_path_step_all(
     def step_all(self):
         logger.info("integrating driver forward in time")
         with self.performance_collector.total_timer.clock("total"):
+            self.profiler.enable()
```
elynnwu, line 602: It doesn't actually make sense to have this for dace orchestration, do we want to support this for other backends?

.jenkins/driver_performance.sh
```diff
@@ -0,0 +1,51 @@
...
+#SBATCH --partition=normal
+########################################################
+set -x
+export OMP_NUM_THREADS=12
+export FV3_DACEMODE=BuildAndRun
+srun python -m pace.driver.run ${JENKINS_DIR}/driver_configs/baroclinic_c192_6ranks.yaml
+EOF
+launch_job run.daint.slurm 3600
+
+cat << EOF > results.py
```
twicki, line 39: this should probably live somewhere and not be created on the fly, right?

driver/pace/driver/driver.py
twicki: I would keep it here

util/pace/util/_profiler.py
```diff
@@ -0,0 +1,44 @@
...
+        self._enabled = False
+        self.profiler.dump_stats(filename)
+
+    @property
+    def enabled(self) -> bool:
+        """Indicates whether the profiler is currently enabled."""
+        return self._enabled
+
+
+class NullProfiler(Profiler):
```
twicki, line 24: why are we inheriting from the base-profiler here? This seems dangerous. If we want a base profiler it should be an abc, right?

twicki (changes requested):

.jenkins/driver_performance.sh
elynnwu: Moved to be a separate file.

util/pace/util/_profiler.py
elynnwu: Removed inheritance

twicki (approved):
lgtm

PR #370, opened 2022-10-28
Feature/diagnostics level select

elynnwu:
## Purpose

This PR adds the capability of saving a vertical slice using `z_select` in the diagnostics config.

The expected syntax is as follows:
```
diagnostics_config:
  path: output
  output_format: netcdf
  z_select:
    - level: 1
       names:
         - ua
         - va
    - level: 10
       names:
         - pt  
```

## Code changes:

- diagnostics.py: added `ZSelect` class for storing selected vertical slice and the variables to be saved
   - fix column integral mistake, need to multiple by `1/gravity`

## Checklist


driver/pace/driver/diagnostics.py
```diff
@@ -34,6 +36,35 @@ def cleanup(self):
         ...
 
 
+@dataclasses.dataclass
+class ZSelect:
+    level: int
+    names: List[str]
+
+    def slice_data(self, state: DycoreState):
```
twicki, line 44: technically we select and don't slice here - not super relevant as selecting is a 1-sized slice, but if we actually wan to implement both we might want to rename it here already?

driver/pace/driver/diagnostics.py
```diff
@@ -34,6 +36,35 @@ def cleanup(self):
         ...
 
 
+@dataclasses.dataclass
+class ZSelect:
+    level: int
```
twicki, line 41: what is the main motivation behind 1 level for all names?

driver/pace/driver/diagnostics.py
```diff
@@ -34,6 +36,35 @@ def cleanup(self):
...
+        output = {}
+        for name in self.names:
+            if name not in state.__dict__.keys():
+                raise ValueError(f"Invalid state variable {name} for level select")
+            assert len(getattr(state, name).dims) > 2
+            if getattr(state, name).dims[2] != (
+                pace.util.Z_DIM or pace.util.Z_INTERFACE_DIM
+            ):
+                raise ValueError(
+                    f"z_select only works for state variables with a z dimension. \
```
twicki, line 54: we don't actually have a field that triggers this, right?

twicki (approved):
just a few questions

driver/pace/driver/diagnostics.py
elynnwu: Changed it to select data.

driver/pace/driver/diagnostics.py
elynnwu: That's the more intuitive way for me. Also, ZSelect implies that we are selecting by z level. So I think this makes sense.

driver/pace/driver/diagnostics.py
elynnwu: Correct, while writing test I realized this is not possible based on the current DycoreState.

elynnwu: Added test

PR #375, opened 2022-11-07
Update gt4py on 2022.11.07

jdahm:
## Purpose

Updates the gt4py submodule to a newer version.

## Code changes

This removes the `storage` attribute from `Quantity`.


jdahm:
This requires some fixes which are much easier after https://github.com/ai2cm/pace/pull/377 is merged.

dsl/pace/dsl/dace/orchestration.py
```diff
@@ -36,16 +36,23 @@
...
 
 def _upload_to_device(host_data: List[Any]):
     """Make sure any data that are still a gt4py.storage gets uploaded to device"""
-    for data in host_data:
-        if isinstance(data, gt4py.storage.Storage):
-            data.host_to_device()
+    assert cp is not None
+    for i, data in enumerate(host_data):
+        if isinstance(data, cp.ndarray):
+            host_data[i] = cp.asarray(data)
```
jdahm, line 55: This might need to work in both cases: whether it's a CPU or GPU backend.

stencils/pace/stencils/testing/translate.py
```diff
@@ -204,15 +204,11 @@ def slice_output(self, inputs, out_data=None):
                 )
                 for varname, data_element in data_result.items():
                     index = names_4d.index(varname)
-                    if hasattr(data_element, "synchronize"):
-                        data_element.synchronize()
```
jdahm, line 208: I believe these were unnecessary, but not sure. There is already a call `utils.device_sync(backend=self.stencil_factory.backend)` above this.

util/pace/util/initialization/allocator.py
```diff
@@ -91,9 +101,11 @@ def from_array(
...
         for the given dims.
         """
-        base = self.empty(dims=dims, units=units, dtype=data.dtype)
-        base.data[:] = base.np.asarray(data)
-        return base
+        # TODO: Replace this once aligned_index fix is included.
+        quantity_data = self._numpy.from_array(
+            data, data.dtype, aligned_index=[0] * len(data.shape)
+        )
+        return Quantity(data=quantity_data, dims=dims, units=units)
```
jdahm, line 108: This is a rather large change, but something like this is required because `base.data[:] = base.np.asarray(data)` doesn't work.

util/pace/util/quantity.py
```diff
@@ -489,20 +444,18 @@ def np(self) -> NumpyModule:
...
     @property
     def shape(self):
-        return self.storage.shape
+        return self.data.shape
 
     def __descriptor__(self):
-        if self._storage is None:
-            return None  # trigger DaCe JIT
-        return self._storage.__descriptor__()
+        return None  # trigger DaCe JIT
```
jdahm, line 458: Should get clearer on this before merging too.

constraints.txt
```diff
@@ -38,7 +38,7 @@ babel==2.9.1
     # via sphinx
 backcall==0.2.0
     # via ipython
-backports.entry-points-selectable==1.1.1
+backports-entry-points-selectable==1.1.1
```
mcgibbon, line 41: Any idea why these changed?

driver/pace/driver/driver.py
```diff
@@ -595,9 +595,9 @@ def _critical_path_step_all(
                     self.end_of_step_update(
                         dycore_state=self.state.dycore_state,
                         phy_state=self.state.physics_state,
-                        u_dt=self.state.tendency_state.u_dt.storage,
-                        v_dt=self.state.tendency_state.v_dt.storage,
-                        pt_dt=self.state.tendency_state.pt_dt.storage,
+                        u_dt=self.state.tendency_state.u_dt.data,
+                        v_dt=self.state.tendency_state.v_dt.data,
+                        pt_dt=self.state.tendency_state.pt_dt.data,
                         dt=float(dt),
```
mcgibbon, line 601: Quantity should work here.
```suggestion
                        u_dt=self.state.tendency_state.u_dt,
                        v_dt=self.state.tendency_state.v_dt,
                        pt_dt=self.state.tendency_state.pt_dt,
                        dt=float(dt),
```

dsl/pace/dsl/dace/orchestration.py
mcgibbon: Yeah, I think this changes the code in a way that will break it. I do wonder though whether this function is needed at all. It's called only when parsing/calling dace programs, to convert input arguments to gpu data to pass to dace. However, all storage data should already be allocated on gpu in that case, so this call should be redundant. This is evidenced by it doing nothing to Quantity, and all arguments to the dace programs should be Quantity, not Storage.

dsl/pace/dsl/dace/orchestration.py
```diff
@@ -55,9 +62,7 @@ def _download_results_from_dace(
     gt4py_results = None
     if dace_result is not None:
         for arg in args:
-            if isinstance(arg, gt4py.storage.Storage) and hasattr(
-                arg, "_set_device_modified"
-            ):
+            if isinstance(arg, cp.ndarray) and hasattr(arg, "_set_device_modified"):
```
mcgibbon, line 65: Also wondering here what happens to Quantity.

Can you use a `try: except` block that calls this method instead of isinstance checks?

dsl/pace/dsl/gt4py_utils.py
```diff
@@ -401,22 +419,16 @@ def index(array, key):
 
 
 def moveaxis(array, source: int, destination: int):
```
mcgibbon, line 421: Not for this PR, but we should probably try to delete these functions, in favor of e.g. `quantity.np.moveaxis`.

dsl/pace/dsl/stencil.py
```diff
@@ -32,13 +31,16 @@
...
+    cp = np
+
+
 def report_difference(args, kwargs, args_copy, kwargs_copy, function_name, gt_id):
     report_head = f"comparing against numpy for func {function_name}, gt_id {gt_id}:"
     report_segments = []
     for i, (arg, numpy_arg) in enumerate(zip(args, args_copy)):
-        if isinstance(arg, pace.util.Quantity):
-            arg = arg.storage
-            numpy_arg = numpy_arg.storage
```
mcgibbon, line 41: Why is this block removed? It should still be needed?

dsl/pace/dsl/stencil.py
```diff
@@ -518,15 +520,11 @@ def closure_resolver(self, constant_args, given_args, parent_closure=None):
...
-            pass
+        if isinstance(arg, pace.util.Quantity):
+            args[i] = arg.data
     for name, arg in kwargs.items():
-        try:
-            kwargs[name] = arg.storage
-        except AttributeError:
-            pass
+        if isinstance(arg, pace.util.Quantity):
+            kwargs[name] = arg.data
```
mcgibbon, line 527: Please use `try: except` as before instead of `isinstance`

physics/tests/savepoint/translate/translate_fv_update_phys.py
```diff
@@ -197,9 +203,9 @@ def compute_parallel(self, inputs, communicator):
         state.v = v_quantity
         self._base.compute_func(
             state,
-            tendencies["u_dt"].storage,
-            tendencies["v_dt"].storage,
-            tendencies["t_dt"].storage,
+            tendencies["u_dt"].data,
+            tendencies["v_dt"].data,
+            tendencies["t_dt"].data,
```
mcgibbon, line 208:
```suggestion
            tendencies["u_dt"],
            tendencies["v_dt"],
            tendencies["t_dt"],
```

stencils/pace/stencils/testing/translate.py
```diff
@@ -25,6 +31,17 @@ def pad_field_in_j(field, nj: int, backend: str):
     return outfield
 
 
+def get_data(value: Union[pace.util.Quantity, np.ndarray]) -> np.ndarray:
```
mcgibbon, line 34: nit: Can you name this `as_numpy` or similar?

stencils/pace/stencils/testing/translate.py
mcgibbon: Sounds likely.

util/pace/util/initialization/allocator.py
mcgibbon: This looks dangerous, will slack to clarify.

util/pace/util/quantity.py
```diff
@@ -387,44 +370,16 @@ def sel(self, **kwargs: Union[slice, int]) -> np.ndarray:
...
-    def _initialize_storage(self, data, origin, gt4py_backend: str, mask: Tuple):
-        storage = gt4py.storage.storage.empty(
-            gt4py_backend,
-            default_origin=origin,
-            shape=data.shape,
-            dtype=data.dtype,
-            mask=mask,
-            managed_memory=True,  # required to get GPUStorage with only gpu data copy
+    def _initialize_data(self, data, origin, gt4py_backend: str, dimensions: Tuple):
+        """Allocates an ndarray with optimal memory layout, and copies the data over."""
```
mcgibbon, line 374: This is fine for now. It would be nice if this can avoid allocating new memory if the ndarray is already in the optimal memory layout, or to be able to avoid copy while initializing a Quantity with a non-optimal memory layout (e.g. for NOAA's case).

util/pace/util/quantity.py
mcgibbon: +1

util/tests/quantity/test_storage.py
```diff
@@ -189,37 +136,4 @@ def test_numpy_data_becomes_cupy_with_gpu_backend(
         units=units,
         gt4py_backend=gt4py_backend,
     )
-    assert isinstance(quantity.data, cupy.ndarray)
-    assert isinstance(quantity.storage, gt4py.storage.storage.GPUStorage)
-
-
-@pytest.mark.parametrize("backend", ["gt4py_numpy"], indirect=True)
```
mcgibbon, line 196: I guess if you pass a cupy array on a cpu backend it just performs a data copy and vice-versa? If so that should be fine.

util/pace/util/initialization/allocator.py
mcgibbon: I think you're missing a `gt4py_backend` argument here, as well as origin and extent. This is normally handled automatically by `self.empty`. I'm also not sure whether the resulting data will or won't be optimized for the gt4py backend.

Why does `base.data[:] = base.np.asarray(data)` not work?

dsl/pace/dsl/stencil.py
jdahm: In this case that doesn't work because both Quantity and ndarrays have a data attribute

jdahm: We could rename the data attribute for now to disambiguate, and/or continue down the path of making Quantity an ndarray instead of holding one.

dsl/pace/dsl/stencil.py
jdahm: Good catch. I'll add it back with `arg.data` instead of `arg.storage`.

stencils/pace/stencils/testing/translate.py
jdahm: Sure!

dsl/pace/dsl/stencil.py
mcgibbon: Given where this code is, it's performance critical to use a try:except block here over isinstance. Can you add a do-nothing line that just attempts to access an attribute we know only exists on Quantity, and add a `# code comment` explaining why it's there?

constraints.txt
jdahm: No, and a quick Google search tells me something is probably amiss here. Will investigate.

dsl/pace/dsl/gt4py_utils.py
jdahm: Good idea, I added an issue [here](https://github.com/ai2cm/pace/issues/383).

dsl/pace/dsl/dace/orchestration.py
jdahm: I think I fixed this by looking at all references and checking for `config.is_gpu_backend()` before calling. I'll also add a note here that this will raise an exception if it is called otherwise.

constraints.txt
jdahm: For now I just set this back.

util/pace/util/quantity.py
jdahm: I made a change that checks to see if the data is in the "optimal" layout and skips this if it is. Nice catch!

util/pace/util/initialization/allocator.py
jdahm: Just tried it. This happens:
```     
        base = self.empty(dims=dims, units=units, dtype=data.dtype)
>       base.data[:] = base.np.asarray(data)
E       ValueError: could not broadcast input array from shape (19,19,80) into shape (19,19,3)
```

mcgibbon: That's a fairly serious problem worth debugging. Why was `base` initialized by `self.empty` with a z-axis length of 3? That shouldn't happen.

dsl/pace/dsl/stencil.py
jdahm: Ok I added that.

util/pace/util/initialization/allocator.py
jdahm: Thanks for helping resolve this. We now have the arrays clipped to the correct shape.

util/pace/util/quantity.py
```diff
@@ -476,14 +476,7 @@ def __descriptor__(self) -> "dace.data.Array":
...
             raise ModuleNotFoundError("dace is not installed")
-        storage = (
-            dace.StorageType.GPU_Global
-            if hasattr(self.data, "__cuda_array_interface__")
-            else dace.StorageType.CPU_Heap
-        )
-        return dace.data.Array(
-            dace.typeclass(str(self.data.dtype)), self.shape, storage=storage
-        )
+        dace.data.create_datadescriptor(self.data)
```
gronerl, line 479: I just happened to look at this commit by chance, I think it needs a `return` ;)

jdahm: Yes, thanks!

util/pace/util/initialization/allocator.py
```diff
@@ -50,6 +50,12 @@ def from_backend(cls, sizer: GridSizer, backend: str):
         numpy = StorageNumpy(backend)
         return cls(sizer, numpy)
 
+    def _backend(self) -> Optional[str]:
+        if hasattr(self._numpy, "backend"):
+            return self._numpy.backend
+        else:
+            return None
```
jdahm, line 57:
```suggestion
        try:
            return self._numpy.backend
        except AttributeError:
            return None
```

FlorianDeconinck (approved):
Johann and I co-author and cross reviewed each other

gronerl:
:tada:

PR #379, opened 2022-11-08
add safety checks for guarantees on the state during runs

twicki:
## Purpose

As per discussion we want a way to test that our state is still sane during runs

## Code changes:

- Added a `SafetyChecker` in the main driver that is called every `safety_check_frequency` in the `_end_of_step_actions` and checks the state for consistency
- Added optional  `safety_check_frequency` to the DriverConfig
- Added `SafetyChecker` that does check on local data
- Added tests for the `SafetyChecker`

## Checklist

### DIFF
--- a/driver/pace/driver/driver.py
+++ b/driver/pace/driver/driver.py
@@ -518,7 +518,7 @@ class Driver:
         self._time_run = self.config.start_time
 
         self.safety_checker = SafetyChecker()
-        SafetyChecker.register_variable("ua", -200, 200)
+        SafetyChecker.register_variable("ua", -200, 200, compute_domain_only=True)
 
     def _update_driver_config_with_communicator(
         self, communicator: CubedSphereCommunicator
--- a/driver/pace/driver/driver.py
+++ b/driver/pace/driver/driver.py
@@ -17,6 +17,7 @@ import pace.stencils
 import pace.util
 import pace.util.grid
 from pace import fv3core
+from pace.driver.safety_checks import SafetyChecker
 from pace.dsl.dace.dace_config import DaceConfig
 from pace.dsl.dace.orchestration import dace_inhibitor, orchestrate
 from pace.dsl.stencil_config import CompilationConfig, RunMode
@@ -122,6 +123,7 @@ class DriverConfig:
     pair_debug: bool = False
     output_initial_state: bool = False
     output_frequency: int = 1
+    safety_check_frequency: int = 1
 
     @functools.cached_property
     def timestep(self) -> timedelta:
@@ -515,6 +517,9 @@ class Driver:
 
         self._time_run = self.config.start_time
 
+        self.safety_checker = SafetyChecker()
+        SafetyChecker.register_variable("ua", -200, 200)
+
     def _update_driver_config_with_communicator(
         self, communicator: CubedSphereCommunicator
     ) -> None:
@@ -554,6 +559,8 @@ class Driver:
         self.time += self.config.timestep
         if ((step + 1) % self.config.output_frequency) == 0:
             self.diagnostics.store(time=self.time, state=self.state)
+        if ((step + 1) % self.config.safety_check_frequency) == 0:
+            self.safety_checker.check_state(self.state.dycore_state)
         self.config.restart_config.write_intermediate_if_enabled(
             state=self.state,
             step=step,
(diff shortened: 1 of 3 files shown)


elynnwu:
launch jenkins

driver/pace/driver/driver.py
```diff
@@ -122,6 +123,7 @@ class DriverConfig:
     pair_debug: bool = False
     output_initial_state: bool = False
     output_frequency: int = 1
+    safety_check_frequency: int = 1
```
elynnwu, line 126: I think by default we should not do safety check, I would suggest we set this to a very large number, or we should even have a switch to turn on/off?

driver/pace/driver/driver.py
```diff
@@ -515,6 +517,9 @@ def exit_instead_of_build(self):
 
         self._time_run = self.config.start_time
 
+        self.safety_checker = SafetyChecker()
+        SafetyChecker.register_variable("ua", -200, 200, compute_domain_only=True)
+
```
elynnwu, line 522: I think we should register some variables by default when you make `SafetyChecker`, then we can also allow custom checkers from the namelist

driver/pace/driver/safety_checks.py
```diff
@@ -0,0 +1,89 @@
...
+            raise NotImplementedError("Can only register variables once")
+        else:
+            cls.checks[name] = (minimum_value, maximum_value, compute_domain_only)
+
+    @classmethod
+    def clear_all_checks(cls):
+        """Clear all the registered checks"""
+        cls.checks.clear()
+
+    def check_state(self, state: DycoreState):
```
elynnwu, line 50: Allow this for `PhysicsState` also

driver/pace/driver/safety_checks.py
```diff
@@ -0,0 +1,89 @@
...
+                min_value = var.data[
+                    var.origin[0] : var.origin[0] + var.extent[0],
+                    var.origin[1] : var.origin[1] + var.extent[1],
+                    var.origin[2] : var.origin[2] + var.extent[2],
+                ].min()
+                max_value = var.data[
+                    var.origin[0] : var.origin[0] + var.extent[0],
+                    var.origin[1] : var.origin[1] + var.extent[1],
+                    var.origin[2] : var.origin[2] + var.extent[2],
+                ].max()
```
elynnwu, line 75: I believe every variables in dycore and physics state are now quantity, so you can do `quantity.view` for compute domain

driver/pace/driver/safety_checks.py
```diff
@@ -0,0 +1,89 @@
...
+        Raises:
+            NotImplementedError: If one of the registered variables are not in the state
+            RuntimeError: If one of the variables exceeds its specified bounds
+        """
+        for variable, bounds in self.checks.items():
+            try:
+                var: Quantity = state.__getattribute__(variable)
+            except AttributeError:
+                raise NotImplementedError("Variable is not in the state")
+            if bounds[2]:
```
elynnwu, line 65: What is `bounds[2]`?

driver/pace/driver/safety_checks.py
```diff
@@ -0,0 +1,89 @@
...
+            if bounds[0] and min_value < bounds[0]:
+                raise RuntimeError(
+                    f"Variable {variable} is outside of its specified bounds: \
+                    {bounds[0]} specified, {min_value} found"
+                )
+            if bounds[1] and max_value > bounds[1]:
+                raise RuntimeError(
+                    f"Variable {variable} is outside of its specified bounds: \
+                    {bounds[1]} specified, {max_value} found"
+                )
```
elynnwu, line 89: Do you know if it errors out here, it will still go and write out a restart file? I think it would be helpful to dump some data when safety check is on and an out of bounds error is detected.

twicki: This happens in the end-of step, so it will be caught by the try / finally block, so if output is specified, it will dump, yes

driver/pace/driver/safety_checks.py
twicki: the check of if domain-size only or on the entire domain. Renamed for clarifiaction

driver/pace/driver/safety_checks.py
twicki: nice, makes life easier

driver/pace/driver/safety_checks.py
twicki: I don't think this is a good idea. Variables that have the same name in both places are an issue

PR #392, opened 2022-11-23
Update gh-pages with a circleci job

jdahm:
## Purpose

Updates the documentation on the `gh-pages` for every merge to `main`.

TODO:
- [x] Check that this works
- [x] Add credentials
- [x] Ensure it only executes on commits on `main`

## Infrastructure changes:

- Adds a new CircleCI job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -495,8 +495,8 @@ jobs:
             rm -rf gh-pages/docs
             mv docs/_build/html gh-pages/docs
             hash=$(git rev-parse --short HEAD)
-            git config user.name "CircleCI"
-            git config user.email "ai2cm@allenai.org"
+            git config --global user.name "CircleCI"
+            git config --global user.email "ai2cm@allenai.org"
             git -C gh-pages commit -a -m "Docs for $hash"
             git -C gh-pages push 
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -469,6 +469,39 @@ jobs:
...
+            pip install -r requirements_docs.txt -c constraints.txt
+      - run:
+          name: Make docs
+          command: |
+            . venv/bin/activate
+            make docs
+      - run:
+          name: Move into place
+          command: |
+            git clone --branch gh-pages git@github.com:ai2cm/pace.git gh-pages
```
mcgibbon, line 494: Can you merge this plan with the doctest plan? These should only be pushed if doctests pass anyways.

jdahm: As far as I can tell these two plans are doing different things, and should be run at different times, so I'm not sure what commonality there is to merge them. The new plan `push_docs` builds and pushes docs when a PR is merged to `main`, while the `doctest` plan tests that the sphinx docs build and the snippets run without error.

jdahm: Right, this will only run if the `doctest` plan passes, since that is a branch protection requirement.

elynnwu (approved):
lgtm

PR #405, opened 2022-12-14
Update to gt4py v1.0

jdahm:
## Purpose

Updates to the latest gt4py that includes the code reorganization, so this updates those import paths.


constraints.txt
```diff
@@ -38,7 +38,7 @@ babel==2.9.1
     # via sphinx
 backcall==0.2.0
     # via ipython
-backports.entry-points-selectable==1.1.1
+backports-entry-points-selectable==1.1.1
```
jdahm, line 41:
```suggestion
backports.entry-points-selectable==1.1.1
```

PR #418, opened 2022-12-22
Fix notebook image and update for no storage

jdahm:
## Purpose

- Adds a non-editable install of the pace modules in the Docker build.
- Removes the `default_origin` keyword argument in one call to `gt4py.storage.zeros` in an example notebook.


mcgibbon (approved):
LGTM. May want to add a comment describing why the non-edit installs were needed.

